PR #12, opened 2015-04-03
Landing strip

ouchadam:
A library for tab strips which relies on the client to do all the hard work

Based on @jpardogo 's https://github.com/jpardogo/PagerSlidingTabStrip which is in turn a fork from @astuetz https://github.com/astuetz/PagerSlidingTabStrip (mad props to those guys!)

The design difference here is the `LandingStrip` view is now composed of a tab layout by default, whereas originally the `Strip` view controlled all styling and had the unfortunate sideeffect of overriding any custom styles. 

This means the `LandingStrip` view has been heavily simplified and allows for complete customisation (because the library leaves all that hard work to the client).

The current state of the lib is a barebones working implementation for our(/my) needs in another project. The api, name etc have not been fully defined yet.

ezgif com-video-to-gif


mr-archano:
It would be good to understand more about the purpose of this fork/rewrite. The diff contains too much noise, it's hard to tell what one should focus on.

ouchadam:
this file is the library https://github.com/novoda/spikes/blob/landing-strip/landing-strip/core/src/main/java/com/novoda/landingstrip/LandingStrip.java the rest of the project is examples and build script setup. I'll update the description with the reasonings

landing-strip/core/src/main/java/com/novoda/landingstrip/LandingStrip.java
```diff
@@ -0,0 +1,289 @@
...
+import android.support.v4.view.PagerAdapter;
+import android.support.v4.view.ViewPager;
+import android.util.AttributeSet;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.widget.HorizontalScrollView;
+import android.widget.LinearLayout;
+import android.widget.TextView;
+
+public class LandingStrip extends HorizontalScrollView {
```
ouchadam: This is the meat of the lib

mr-archano:
Great description @ouchadam :+1: Also the link to the diff between libs is pure gold :sparkles:.

landing-strip/core/src/main/java/com/novoda/landingstrip/Attributes.java
```diff
@@ -0,0 +1,87 @@
...
+import android.content.res.TypedArray;
+import android.support.annotation.ColorRes;
+import android.support.annotation.LayoutRes;
+import android.util.AttributeSet;
+
+class Attributes {
+
+    @ColorRes
+    private static final int DEFAULT_INDICATOR_COLOUR = android.R.color.white;
+    private static final int DEFAULT_INDICATOR_HEIGHT = 5;
```
ataulm: this should be `@DimenRes` and point to a dimen - if it's default, it should be default DPs not px

landing-strip/core/src/main/java/com/novoda/landingstrip/Attributes.java
```diff
@@ -0,0 +1,87 @@
...
+import android.content.Context;
+import android.content.res.TypedArray;
+import android.support.annotation.ColorRes;
+import android.support.annotation.LayoutRes;
+import android.util.AttributeSet;
+
+class Attributes {
+
+    @ColorRes
+    private static final int DEFAULT_INDICATOR_COLOUR = android.R.color.white;
```
ataulm: `android.R.color` can be different on different platforms / manufacturer's versions - this should be an internal to the project default.

landing-strip/core/src/main/java/com/novoda/landingstrip/Attributes.java
```diff
@@ -0,0 +1,87 @@
...
+        } finally {
+            xml.recycle();
+        }
+    }
+
+    private static void throwIfTabLayoutIsMissing(int tabLayoutId) {
+        if (tabLayoutId != MISSING_TAB_LAYOUT_ID) {
+            return;
+        }
+        throw new MissingTabLayoutId();
```
ataulm: `new MissingTabLayoutId**Exception**()`

landing-strip/core/src/main/java/com/novoda/landingstrip/Attributes.java
```diff
@@ -0,0 +1,87 @@
...
+        this.indicatorHeight = indicatorHeight;
+        this.tabsPaddingLeft = tabsPaddingLeft;
+        this.tabsPaddingRight = tabsPaddingRight;
+    }
+
+    public int getTabLayoutId() {
+        return tabLayoutId;
+    }
+
+    public int getIndicatorColour() {
```
ataulm: `color` por favor. It's just weird for one library out of the entire eco system of android libraries and android APIs to use UK english - different (but still gross!) for private APIs

landing-strip/core/src/main/java/com/novoda/landingstrip/Attributes.java
```diff
@@ -0,0 +1,87 @@
...
+        return indicatorColour;
+    }
+
+    public int getIndicatorHeight() {
+        return indicatorHeight;
+    }
+
+    public int getTabsPaddingLeft() {
```
ataulm: from the client point of view won't this just be `getPaddingLeft()` (and `..Right()`)? or is this the padding that's applied to each individual tab?

landing-strip/core/src/main/java/com/novoda/landingstrip/Coordinates.java
```diff
@@ -0,0 +1,6 @@
+package com.novoda.landingstrip;
+
+public interface Coordinates {
```
ataulm: yagni?

landing-strip/core/src/main/java/com/novoda/landingstrip/Attributes.java
ouchadam: there are 3 types of paddings
- individual tab (each child view)
- tabs container (the linear layout which contains the tabs)
- the parent view padding (the horizontal screen view itself)

this is meant to indicate the tabs container padding

landing-strip/core/src/main/java/com/novoda/landingstrip/LandingStrip.java
```diff
@@ -0,0 +1,133 @@
...
+        this.attributes = Attributes.readAttributes(context, attrs);
+        this.layoutInflater = LayoutInflater.from(context);
+        this.indicatorPaint = new Paint();
+        this.state = new State();
+        this.tabsContainer = new LinearLayout(context);
+        this.indicatorCoordinatesCalculator = IndicatorCoordinatesCalculator.newInstance();
+
+        state.updateDelegateOnPageListener(new ViewPager.SimpleOnPageChangeListener());
+        state.updatePosition(0);
+        state.updatePositionOffset(0);
```
ataulm: why aren't these done in the ctor of `State`?

landing-strip/core/src/main/java/com/novoda/landingstrip/Attributes.java
ouchadam: :+1:  yepo

landing-strip/core/src/main/java/com/novoda/landingstrip/Attributes.java
ouchadam: done :+1:

landing-strip/core/src/main/java/com/novoda/landingstrip/Coordinates.java
ouchadam: it's because of the api, the clients shouldn't be able to modify the properties, only the lib can do that

landing-strip/core/src/main/java/com/novoda/landingstrip/LandingStrip.java
ouchadam: `ctor` or `newInstance`?

landing-strip/core/src/main/java/com/novoda/landingstrip/Coordinates.java
ouchadam: the coordinates are mutable because we don't want to be new'ing up objects in `onDraw`

landing-strip/core/src/main/java/com/novoda/landingstrip/Attributes.java
ouchadam: very true :+1:

landing-strip/core/src/main/java/com/novoda/landingstrip/Coordinates.java
ataulm: :+1:

landing-strip/core/src/main/java/com/novoda/landingstrip/LandingStrip.java
ataulm: :fried_shrimp: either for me - just not here. I don't follow why position and positionOffset would ever be non-zero when you create a State object.

ouchadam: more for explicitness

ataulm:
:+1: merging to spikes to continue spiking!

xrigau:
:+1: good stuff!

landing-strip/core/src/main/java/com/novoda/landingstrip/LandingStrip.java
ataulm: but you don't do the same with other things - `indicatorPaint.setColor(0)` or `tabsContainer.removeAllViews()`

ouchadam: they aren't part of the state object

ataulm: but you're doing `new State()` :s

PR #13, opened 2015-04-27
Add demo activity for View only pager

ataulm:
We're using a non-fragment setup in a couple of projects where we'd like to use LandingStrip (e.g. Wutson).

This demo serves as a testing ground for Wutson at least, which uses this version of ViewPagerAdapter to make sure that changes on LandingStrip were compatible.


ouchadam:
:100:

xrigau:
:+1:

PR #15, opened 2015-05-19
Bug fix layout inflation using incorrect parent

StylingAndroid:
Fixes a bug in tab layout inflation which now permits fixed width operation:

screenshot-2015-05-19_11 54 31 867


ouchadam:
:100:

PR #17, opened 2015-09-12
Talkback navdrawer

ataulm:
@Electryc @blundell @Dorvaryn 

- We don't navigate on navigation item click for some reason (jankiness)
- We use the `onDrawerClosed` event to then actually navigate.
- Closing the open drawer will trigger an a11y event - it'll announce the drawer has been closed, and then it'll read the now visible (the old one) activity, followed by the new one.

I think this is because the messages are queued - Kelly Shuster mentioned something in her berlin talk about "polite" mode (I forget the details) where you can either mark your accessibility announcements as polite/brutal to wait til it's finished what it's saying or interrupt it. Perhaps it ought to be in brutal mode, but I don't think that's under our control.

### DIFF
--- a/ActivityTalkbackInvestigation/app/src/main/java/com/novoda/activitytalkbackinvestigation/ProHaxDrawerActivity.java
+++ b/ActivityTalkbackInvestigation/app/src/main/java/com/novoda/activitytalkbackinvestigation/ProHaxDrawerActivity.java
@@ -18,6 +18,8 @@ import android.widget.ListView;
 
 public abstract class ProHaxDrawerActivity extends AppCompatActivity {
 
+    private static final String EXTRA_INTERNAL_NAV = BuildConfig.APPLICATION_ID + "EXTRA_INTERNAL_NAV";
+
     private DrawerLayout drawerLayout;
 
     @Override
@@ -26,8 +28,11 @@ public abstract class ProHaxDrawerActivity extends AppCompatActivity {
         ViewGroup contentView = (ViewGroup) findViewById(R.id.content_frame);
         getLayoutInflater().inflate(layoutResID, contentView);
         setupDrawer();
-        drawerLayout.setImportantForAccessibility(View.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS);
-        drawerLayout.openDrawer(GravityCompat.START);
+
+        if (getIntent().getExtras() != null && getIntent().getExtras().containsKey(EXTRA_INTERNAL_NAV)) {
+            drawerLayout.setImportantForAccessibility(View.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS);
+            drawerLayout.openDrawer(GravityCompat.START);
+        }
     }
 
     @Override
@@ -80,11 +85,14 @@ public abstract class ProHaxDrawerActivity extends AppCompatActivity {
     }
 
     private static Intent createIntentForItem(int itemPosition, Context context) {
+        Intent intent;
         if (itemPosition == 0) {
-            return new Intent(context, EarthActivity.class);
+            intent = new Intent(context, EarthActivity.class);
         } else {
-            return new Intent(context, SaturnActivity.class);
+            intent = new Intent(context, SaturnActivity.class);
         }
+        intent.putExtra(EXTRA_INTERNAL_NAV, true);
+        return intent;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


blundell:
so this reproduces the bug or fixes it?

ataulm:
This reproduces it.

To fix the undesired behaviour with TalkBack, just open the activity when the user clicks on the navigation item, instead of waiting for the drawer to close.

But there was a reason we were waiting for the drawer to close - there was jankiness if we tried to open a new activity while the drawer was closing.

I've pushed an alternative NavigationDrawer activity (ProHaxDrawerActivity), showing how we might want to override the default activity transitions when opening an activity _via the navigation drawer_ to make it look more like Inbox, Play Store etc.

This also involves a bit of trickery making it look like the navigation drawer is staying open while the content behind is changing; really, it's a different drawer (:boom:) that's pre-opened. And all the close drawer animations have arbitrary delays. The only thing further to this would be to ensure that the navigation drawer isn't open if the activity was opened from something other than our nav drawer (i.e. the app launcher).

boom webm

ActivityTalkbackInvestigation/app/src/main/java/com/novoda/activitytalkbackinvestigation/ProHaxDrawerActivity.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private DrawerLayout drawerLayout;
+
+    @Override
+    public void setContentView(@LayoutRes int layoutResID) {
+        super.setContentView(R.layout.drawer_layout);
+        ViewGroup contentView = (ViewGroup) findViewById(R.id.content_frame);
+        getLayoutInflater().inflate(layoutResID, contentView);
+        setupDrawer();
+        drawerLayout.openDrawer(GravityCompat.START);
```
ataulm: this would need to be guarded somehow so it doesn't open the drawer if the activity was opened by the app launcher.

ActivityTalkbackInvestigation/app/src/main/java/com/novoda/activitytalkbackinvestigation/ProHaxDrawerActivity.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private DrawerLayout drawerLayout;
+
+    @Override
+    public void setContentView(@LayoutRes int layoutResID) {
+        super.setContentView(R.layout.drawer_layout);
+        ViewGroup contentView = (ViewGroup) findViewById(R.id.content_frame);
+        getLayoutInflater().inflate(layoutResID, contentView);
+        setupDrawer();
+        drawerLayout.setImportantForAccessibility(View.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS);
```
ataulm: The other quick fix (without any of the rest, I suppose) which can be applied is to tell android that the drawer layout is not important for accessibility in `onItemClick`, then set it back to AUTO in `onDrawerClosed`.

ataulm:
@Dorvaryn no "windowIsTranslucent="true"` == no white flashes? :grin:

blundell:
does this still just reproduce it? lol I want to merge it as reproducing - then separate PR for fix

ataulm:
Errr the other activity is unused so yes it reproduces it. The fix would be
using the other base activity.

Don't knoow what to dooo

On Sun, 13 Sep 2015 14:27 Paul Blundell <notifications@github.com> wrote:

> does this still just reproduce it? lol I want to merge it as reproducing -
> then separate PR for fix
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/novoda/spikes/pull/17#issuecomment-139873643>.
>

-- 
Disclaimer: The information in this e-mail and any attachments is the 
property of Novoda Ltd and is confidential and may be legally privileged. 
This e-mail is intended solely for the person or organisation to which it 
is addressed. Any disclosure, copying or other use of the information by 
any person or organisation who is not the intended recipient is strictly 
prohibited and may be unlawful. If you have received this e-mail in error, 
please inform the sender immediately and delete/destroy this e-mail and any 
copies of it. Novoda Ltd has taken reasonable precautions to minimise the 
risk of any software viruses which may damage your systems, but we advise 
that you take the necessary steps to ensure that no virus contamination is 
suffered. Novoda Ltd does not accept any liability for any loss or damage 
caused by the transmission of any virus.

Novoda Ltd, Company No: 347444, Registered in Scotland Registered 
Office: C/O Alexander Sloan, 38 Cadogan Street, Glasgow, G2 7HF, Scotland.  VAT 
Registration Number  GB 984 2525 93

blundell:
:100:

Electryc:
Nice one! :+1:

PR #24, opened 2016-05-11
React/PT-68/Deep Linking

gbasile:
This PR implements: 

- deep-linking functionality required as part of the oauth story.
- Debug Screen, for the moment it contains only the deep-link view
- Basic navigator (swipe from the edge of the screen to go back)

Please have a deep look at this PR as we are implementing new stuff for the first time and probably there are much better ways to accomplish the same. 

Screenshots: 

BEFORE: 

| iOS | Android |
| --- | --- |
| 807064fc-1605-11e6-8cd5-cc8d3c8e97c4 | 889b64b0-1605-11e6-806d-78013a28e50c |

AFTER: 

| iOS | Android |
| --- | --- |
| ios | ezgif com-video-to-gif |

Paired with: 
@eduardb

### DIFF
--- a/ReactNative/ReactTwitter/android/app/src/main/java/com/reacttwitter/OauthIntentPackage.java
+++ b/ReactNative/ReactTwitter/android/app/src/main/java/com/reacttwitter/OauthIntentPackage.java
@@ -1,0 +1,48 @@
+package com.reacttwitter;
+
+import android.net.Uri;
+
+import com.facebook.react.ReactPackage;
+import com.facebook.react.bridge.JavaScriptModule;
+import com.facebook.react.bridge.NativeModule;
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.uimanager.ViewManager;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Created by gbasile on 10/05/2016.
+ */
+public class OauthIntentPackage implements ReactPackage {
+
+    private OauthIntentModule intentModule;
+
+    @Override
+    public List<NativeModule> createNativeModules(ReactApplicationContext reactContext) {
+        List<NativeModule> modules = new ArrayList<>();
+        intentModule = new OauthIntentModule(reactContext);
+        modules.add(intentModule);
+        return modules;
+    }
+
+    @Override
+    public List<Class<? extends JavaScriptModule>> createJSModules() {
+        return Collections.emptyList();
+    }
+
+    @Override
+    public List<ViewManager> createViewManagers(ReactApplicationContext reactContext) {
+        return Collections.emptyList();
+    }
+
+    void handleOnNewIntent(Uri uri) {
+        if (intentModule == null) {
+            return;
+        }
+
+        intentModule.handleOnNewIntent(uri);
+    }
+
+}
--- a/ReactNative/ReactTwitter/android/app/src/main/java/com/reacttwitter/OauthIntentPackage.java
+++ b/ReactNative/ReactTwitter/android/app/src/main/java/com/reacttwitter/OauthIntentPackage.java
@@ -12,9 +12,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-/**
- * Created by gbasile on 10/05/2016.
- */
 public class OauthIntentPackage implements ReactPackage {
 
     private OauthIntentModule intentModule;
--- a/ReactNative/ReactTwitter/android/app/src/main/java/com/reacttwitter/OauthIntentPackage.java
+++ b/ReactNative/ReactTwitter/android/app/src/main/java/com/reacttwitter/OauthIntentPackage.java
@@ -38,10 +38,6 @@ public class OauthIntentPackage implements ReactPackage {
     }
 
     void handleOnNewIntent(Uri uri) {
-        if (intentModule == null) {
-            return;
-        }
-
         intentModule.handleOnNewIntent(uri);
     }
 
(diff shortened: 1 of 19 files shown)


ReactNative/ReactTwitter/core/views/deep-linking.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
+import {
+  StyleSheet,
+  View,
+  Text
+} from 'react-native'
+
+var Button = require('react-native-button')
+var DeepLinkingFacade = require('../service/deep-linking-facade')
```
gbasile: Is there a way to get ride of all these relative path? There is no such thing as a DIR_ROOT variable?

ReactNative/ReactTwitter/core/views/main-navigator.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import {
+  Navigator
+} from 'react-native'
+
+var DebugScreenView = require('./debug-screen.js')
+var DeepLinkingView = require('./deep-linking.js')
+
+const debugScreenID = 'debug-screen-identifier'
```
gbasile: Ideally we would like to define these constants inside our ViewComponents. Any better solutions to avoid these identifier hardcoded here?

ReactNative/ReactTwitter/android/app/src/main/java/com/reacttwitter/OauthIntentPackage.java
```diff
@@ -0,0 +1,44 @@
...
+import com.facebook.react.bridge.NativeModule;
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.uimanager.ViewManager;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Created by gbasile on 10/05/2016.
```
frapontillo: 💣

ReactNative/ReactTwitter/core/service/deep-linking-facade.android.js
```diff
@@ -0,0 +1,21 @@
...
+
+class DeepLinkingFacade {
+
+  static newInstance () {
+    return new DeepLinkingFacade()
+  }
+
+  listenForDeepLinking () {
+    return oAuthIntentAndroid.registerForDeepLinking()
+      .then(function (event) { return event.url })
```
frapontillo: Use lambdas: `(event) => event.url`

ReactNative/ReactTwitter/core/service/deep-linking-facade.ios.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  constructor (linking) {
+    this.linking = linking
+  }
+
+  listenForDeepLinking () {
+    this.linking.removeEventListener('url', this._handleOpenURL)
+    this.linking.addEventListener('url', this._handleOpenURL)
+    return new Promise(function (resolve, reject) {
```
frapontillo: Why are you doing this?

ReactNative/ReactTwitter/core/views/deep-linking.js
frapontillo: You could, but this is how it's done.

ReactNative/ReactTwitter/core/views/main-navigator.js
frapontillo: You could put them in a "constant" module and export them, then get them with `require('../MyConstants').DEBUG_SCREEN_ID`.

ReactNative/ReactTwitter/core/service/deep-linking-facade.ios.js
gbasile: You mean the deep linking?

ReactNative/ReactTwitter/core/views/debug-screen.js
```diff
@@ -0,0 +1,57 @@
...
+        <Button
+          style={styles.button}
+          styleDisabled={styles.button_disabled}
+          onPress={this.pushDeepLinking}>Deep Linking</Button>
+      </View>
+    )
+  },
+
+  pushDeepLinking () {
+    this.props.navigator.push({id: 'deep-linking-identifier'})
```
danybony: 👍

ReactNative/ReactTwitter/core/service/deep-linking-facade.ios.js
frapontillo: No, I mean why are you removing and adding the event listener on every button click? So you can reset the promise?

ReactNative/ReactTwitter/core/views/deep-linking.js
```diff
@@ -0,0 +1,66 @@
...
+
+var DeepLinkingView = React.createClass({
+
+  getInitialState () {
+    console.log('Get initial state')
+    return {
+      facade: DeepLinkingFacade.newInstance(),
+      deepLinkUrl: '[incoming]'
```
danybony: Maybe better to init `deepLinkUrl` to an empty one?

ReactNative/ReactTwitter/core/service/deep-linking-facade.ios.js
gbasile: We wanted to be sure we don't register several times to the event listener. We end up with something that is not really clear, probably we should use a instance variable to know if we are already registered or not

frapontillo: I think this is Rx territory 😛

gbasile: ./francesco --verbose

frapontillo: Promises are for one shot, use-and-discard operations. What you're doing here is starting to listen to events the first time the button is clicked, then do something as soon as events come in.

_Ergo_, this should be an `Observable`, not a `Promise`.

But if this is only for demo reasons I would leave it like this.

PR #25, opened 2016-05-11
resolve class not found issues

devisnik:
Did some tweaking to make it work. 

I moved `BuildProperties` into `BuildPropertiesConfig`, for now only exposing `getInt`. Not sure if this makes sense.

### DIFF
--- a/BuildProperties/buildSrc/src/main/groovy/com/novoda/buildproperties/BuildProperties.groovy
+++ b/BuildProperties/buildSrc/src/main/groovy/com/novoda/buildproperties/BuildProperties.groovy
@@ -14,13 +14,13 @@ class BuildProperties {
     final URI uri;
     final Configuration configuration;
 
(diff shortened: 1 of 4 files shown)


frapontillo:
The purpose of this spike is to make the plugin populate the `project.buildProperties` automatically from one or more properties files.
This is done using a apache commons library.

In your code, though, that library is never called, therefore it's not crashing because it never attempts to call the classes that aren't found.

I have written a decent README that explains what I'm trying to achieve and how to do it. The only issue now seems to be that plugin dependencies aren't resolved by the project applying the plugin.

frapontillo:
Closing this since it turns out it's not something we can solve, it being related to how Apache Commons loads classes (probably reflection): http://stackoverflow.com/questions/37181611/gradle-throws-classnotfoundexception-using-a-plugin-with-dependencies

PR #26, opened 2016-05-12
React/PT-69/Tweets list

danybony:
This PR adds two components shared for both iOS and Android:

- a tweets list component, which in the constructor initialises a mock set of data with a structure similar to the real Twitter data for the home tweet feed
- a single tweet item component, which displays some of the tweet data given a single row passed from the list

Screenshots: 

| iOS | Android |
| --- | --- |
| screen shot 2016-05-11 at 12 35 26 | screen shot 2016-05-11 at 12 35 06 |

Paired with: 
@me


me:
Lies! You didn't pair with @me!

PR #28, opened 2016-05-12
Bonfire / Display message author and time with each message

JozefCeluch:
Displays the author of the message with the date when the message was sent.

bonfire_name_time


dominicfreeston:
That XML! LGTM 👍

PR #30, opened 2016-05-12
Firebase/Add the iOS App

dominicfreeston:
I've naughtily been working in a local repo, so here's the state of work so far:

PT-89/ User can post and receive messages
~~PT-90/ User can write a name in order to identifiy~~
PT-91/ After login user can see whole history
PT-93/ User can authenticate with Google (overrides PT-90)

We've made the decision not to commit Pods, because
a) this is a shared repo, it would be annoying for everyone to have to download dependencies
b) it references Pods that need pre-release credentials, so only properly authorised users should be able to build it

`GoogleService-Info.plist` can be obtained from the Firebase console.

Same as #27 but against `firebase/develop`
Checks have failed because of React spike PRB - we don't actually have any tests 😱 

simulator screen shot 13 may 2016 14 53 23

### DIFF
--- a/Firebase/ios/.gitignore
+++ b/Firebase/ios/.gitignore
@@ -1,0 +1,41 @@
+## Firebase
+GoogleService-Info.plist
+
+## Pods (remove if you want to commit them)
+Pods/
+
+## OS X
+.DS_Store
+
+## Build generated
+build/
+DerivedData
+*.orig
+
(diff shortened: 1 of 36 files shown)


amlcurran:
no screenshots?

dominicfreeston:
screen shot added

PR #35, opened 2016-05-12
Access token Request

eduardb:
We created two helper classes for the Twitter login flow. One that is a generic oauth helper, that creates a signature for a request and builds an authorization header. The second one is is for creating signed request to the Twitter API. For now we only have implemented the `requestToken` method that will provide a token for us to be used with login in the future.

We wrote tests for the `PercentEncoder` and `OauthHelper`, but we had to comment out the later ones because they were failing due to some react native stuff (CryptoJS to be more specific).

To test this functionality, we have included a debug screen accessible from the main screen (Oauth)

| iOS | Android |
| --- | --- |
| may-12-2016 17-36-15 | ezgif com-video-to-gif 1 |

Paired with @gbasile

### DIFF
--- a/ReactNative/ReactTwitter/__tests__/core/service/oauth-helper.test.js
+++ b/ReactNative/ReactTwitter/__tests__/core/service/oauth-helper.test.js
@@ -1,0 +1,38 @@
+jest.unmock('../../../core/service/oauth-helper.js')
+var OauthHelper = require('../../../core/service/oauth-helper.js')
+
+describe('OauthHelper', () => {
+  describe('_collectParameters', () => {
+    it('when collecting parameters', () => {
+      const params = {
+        'status': 'Hello Ladies + Gentlemen, a signed OAuth request!',
+        'include_entities': 'true',
+        'oauth_consumer_key': 'xvz1evFS4wEEPTGEFPHBog',
+        'oauth_nonce': 'kYjzVBB8Y0ZFabxSWbWovY3uYSQ2pTgmZeNu2VS4cg',
+        'oauth_signature_method': 'HMAC-SHA1',
+        'oauth_timestamp': '1318622958',
+        'oauth_token': '370773112-GmHxMAgYyLbNEtIKZeRNFsMKPR9EyMZeS9weJAEb',
+        'oauth_version': '1.0'
+      }
+      expect(OauthHelper._collectParameters(params))
+        .toBe('include_entities=true&oauth_consumer_key=xvz1evFS4wEEPTGEFPHBog&oauth_nonce=kYjzVBB8Y0ZFabxSWbWovY3uYSQ2pTgmZeNu2VS4cg&oauth_signature_method=HMAC-SHA1&oauth_timestamp=1318622958&oauth_token=370773112-GmHxMAgYyLbNEtIKZeRNFsMKPR9EyMZeS9weJAEb&oauth_version=1.0&status=Hello%20Ladies%20%2B%20Gentlemen%2C%20a%20signed%20OAuth%20request%21')
+    })
+  })
+
+  describe('getSignatureBase', () => {
+    it('when getting signature base', () => {
+      const params = {
+        'status': 'Hello Ladies + Gentlemen, a signed OAuth request!',
+        'include_entities': 'true',
+        'oauth_consumer_key': 'xvz1evFS4wEEPTGEFPHBog',
+        'oauth_nonce': 'kYjzVBB8Y0ZFabxSWbWovY3uYSQ2pTgmZeNu2VS4cg',
+        'oauth_signature_method': 'HMAC-SHA1',
+        'oauth_timestamp': '1318622958',
+        'oauth_token': '370773112-GmHxMAgYyLbNEtIKZeRNFsMKPR9EyMZeS9weJAEb',
+        'oauth_version': '1.0'
+      }
+      expect(OauthHelper._getSignatureBase('post', 'https://api.twitter.com/1/statuses/update.json', params))
+        .toBe('POST&https%3A%2F%2Fapi.twitter.com%2F1%2Fstatuses%2Fupdate.json&include_entities%3Dtrue%26oauth_consumer_key%3Dxvz1evFS4wEEPTGEFPHBog%26oauth_nonce%3DkYjzVBB8Y0ZFabxSWbWovY3uYSQ2pTgmZeNu2VS4cg%26oauth_signature_method%3DHMAC-SHA1%26oauth_timestamp%3D1318622958%26oauth_token%3D370773112-GmHxMAgYyLbNEtIKZeRNFsMKPR9EyMZeS9weJAEb%26oauth_version%3D1.0%26status%3DHello%2520Ladies%2520%252B%2520Gentlemen%252C%2520a%2520signed%2520OAuth%2520request%2521')
+    })
+  })
+})
--- a/ReactNative/ReactTwitter/__tests__/core/service/oauth-helper.test.js
+++ b/ReactNative/ReactTwitter/__tests__/core/service/oauth-helper.test.js
@@ -1,38 +1,41 @@
 jest.unmock('../../../core/service/oauth-helper.js')
-var OauthHelper = require('../../../core/service/oauth-helper.js')
+jest.unmock('../../../core/service/percent-encoder.js')
+// var OauthHelper = require('../../../core/service/oauth-helper.js')
 
 describe('OauthHelper', () => {
-  describe('_collectParameters', () => {
-    it('when collecting parameters', () => {
-      const params = {
-        'status': 'Hello Ladies + Gentlemen, a signed OAuth request!',
-        'include_entities': 'true',
-        'oauth_consumer_key': 'xvz1evFS4wEEPTGEFPHBog',
-        'oauth_nonce': 'kYjzVBB8Y0ZFabxSWbWovY3uYSQ2pTgmZeNu2VS4cg',
-        'oauth_signature_method': 'HMAC-SHA1',
-        'oauth_timestamp': '1318622958',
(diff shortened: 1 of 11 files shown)


ReactNative/ReactTwitter/core/service/oauth-helper.js
```diff
@@ -0,0 +1,75 @@
...
+    return output
+  }
+
+  getSigningKey (method, url, params, oauthTokenSecret) {
+    let signingKey = PercentEncoder.encode(this.consumerSecret) + '&' + PercentEncoder.encode(oauthTokenSecret)
+    let signatureBase = OauthHelper._getSignatureBase(method, url, params)
+    return CryptoJS.enc.Base64.stringify(HmacSHA1(signatureBase, signingKey))
+  }
+
+  static _getSignatureBase (method, url, params) {
```
danybony: Is this static for the tests?

eduardb: To be honest, I wanted to avoid instantiating the class for the test, so yes 👅 
But either way, it's not using any object fields, so I don't see a reason why this shouldn't be static. Is it a bad practice?

ReactNative/ReactTwitter/core/service/percent-encoder.js
```diff
@@ -0,0 +1,7 @@
+class PercentEncoder {
```
danybony: What about `WebSafeEncoder`?

ReactNative/ReactTwitter/core/service/oauth-helper.js
```diff
@@ -0,0 +1,75 @@
...
+    return encodedParams
+  }
+
+  static _sortEncodedParams (params) {
+    params.sort((first, second) => {
+      return (first.key > second.key) ? 1 : ((second.key > first.key) ? -1 : 0)
+    })
+  }
+
+  static _joinParams (params) {
```
danybony: Is it a Javascript convention to have a method used only in the current class starting with `_`?

ReactNative/ReactTwitter/core/service/percent-encoder.js
eduardb: I used the terminology from the twitter dev docs here: https://dev.twitter.com/oauth/overview/percent-encoding-parameters

ReactNative/ReactTwitter/core/service/oauth-helper.js
eduardb: No idea, I remember that I saw that somewhere, and that's why I used it this way. Maybe @frapontillo can tell us better?

ReactNative/ReactTwitter/core/views/oauth-screen.js
```diff
@@ -0,0 +1,63 @@
...
+
+  render () {
+    return (
+      <View style={styles.container}>
+      <Button
+        style={styles.button}
+        styleDisabled={styles.button_disabled}
+        onPress={this._buttonClicked}> Request Access Token </Button>
+        <Text style={styles.normal} numberOfLines={2}>
+          Access Token: {"\n"}{this.state.accessToken}
```
danybony: Do you need to pass the `\n` as a thing to be evaluated?

eduardb: No, it's to break the line

ReactNative/ReactTwitter/core/service/oauth-helper.js
danybony: I don't know in Javascript, in general for OOP it's better to have instances of classes wit their specific methods

xrigau:
👏

ReactNative/ReactTwitter/__tests__/core/service/oauth-helper.test.js
```diff
@@ -0,0 +1,41 @@
+jest.unmock('../../../core/service/oauth-helper.js')
+jest.unmock('../../../core/service/percent-encoder.js')
+// var OauthHelper = require('../../../core/service/oauth-helper.js')
+
+describe('OauthHelper', () => {
+
+  // commented out because it's failing for a strange reason
```
xrigau: ?? why not just remove all this commented code?

eduardb: Maybe someday someone will be able to figure out how to fix the tests 💀

ReactNative/ReactTwitter/__tests__/core/service/percent-encoder.test.js
```diff
@@ -0,0 +1,17 @@
+jest.unmock('../../../core/service/percent-encoder.js')
+var PercentEncoder = require('../../../core/service/percent-encoder.js')
+
+describe('PercentEncoder', () => {
+  describe('encode', () => {
+    it('when encoding', () => {
```
frapontillo: You should really use the proper jasmine way of doing this. The pattern is:

```
it('should encode a special character', () => {
    // ...
})

it('should encode only special characters', () => {
    // ...
})

// etc
```

Don't write one `it` for multiple tests, or it'll break badly when (not if 👅) your code fails.

ReactNative/ReactTwitter/core/service/oauth-helper.js
frapontillo: I'm fairly new to ES6, or ES6 is fairly new to me, so I can't give a well formed opinion on this, but I'm good if we are going with the Java convention of making static any method that doesn't use any instance field. 👍

ReactNative/ReactTwitter/core/service/oauth-helper.js
frapontillo: It's a convention to mark private methods with an underscore, yes.

ReactNative/ReactTwitter/core/service/twitter-requests-service.js
```diff
@@ -0,0 +1,51 @@
+var OauthHelper = require('../service/oauth-helper.js')
+
+const CONSUMER_KEY = 'aqPSTs1FT2ndP7qXi247BtbXd'
```
frapontillo: I'd load these from a `config.json`, it's easy as doing `const config = require('path/to/config.json')`

ReactNative/ReactTwitter/core/service/twitter-requests-service.js
```diff
@@ -0,0 +1,51 @@
...
+    const params = {
+      'oauth_callback': CALLBACK_URL,
+      'oauth_consumer_key': CONSUMER_KEY,
+      'oauth_nonce': nonce,
+      'oauth_signature_method': 'HMAC-SHA1',
+      'oauth_timestamp': time,
+      'oauth_version': '1.0'
+    }
+
+    return fetch(url, // eslint-disable-line no-undef
```
frapontillo: You shouldn't use globals, as they are unmockable. Wrap it in your own class if you need to.

ReactNative/ReactTwitter/core/service/twitter-requests-service.js
```diff
@@ -0,0 +1,51 @@
...
+      {
+        method: 'POST',
+        headers: {
+          'Authorization': ouathHelper.buildAuthorizationHeader('post', url, params, '')
+        }
+      })
+    .then((response) => { return response.text() })
+    .then((text) => { return this._getTokenDateFromResponse(text) })
+    .catch((error) => {
```
frapontillo: Or `catch(console.error)` 😉

ReactNative/ReactTwitter/__tests__/core/service/oauth-helper.test.js
frapontillo: We can pair tomorrow if you need some help on this (should be pretty straightforward).

frapontillo:
I left a bunch of comments, feel free to address anyone you agree with, accordingly with your time availability.

eduardb:
I addressed the comments. @frapontillo I don't feel that the tests that I commented out are a priority at this exact moment. Maybe I will get back to them as soon as we finish Twitter login. If you could help then figuring out what crashes, that would be great :)

frapontillo:
👍 for me, good work!

ReactNative/ReactTwitter/core/service/oauth-helper.js
xrigau: also consider if it makes sense and if it makes it easier to read the code, don't just stick with it because it's a JS convention. Same way as we don't use mHungarian notation in android 🔍

ReactNative/ReactTwitter/core/service/oauth-helper.js
xrigau: https://github.com/novoda/spikes/pull/35/files/2fcf0e570834fdf3ca86989f74b44a4b020f5ab3#r63572724

ReactNative/ReactTwitter/__tests__/core/service/oauth-helper.test.js
xrigau: did this pairing happen?

xrigau:
👍

ReactNative/ReactTwitter/core/service/oauth-helper.js
eduardb: There is no other way to tell if a method is private or not, and for me it's very readable...

ReactNative/ReactTwitter/__tests__/core/service/oauth-helper.test.js
eduardb: Uhm, no. Maybe when Francesco comes back from I/O

ReactNative/ReactTwitter/core/service/oauth-helper.js
xrigau: 👍

PR #36, opened 2016-05-12
Firebase / Display public channels in a list

JozefCeluch:
Adds an activity that displays list of all public channels (currently only "global") and allows the user to select a channel.

##### Screenshots

| Before | After |  
| --- | --- |
| channels_before |  channels_after |

Paired with
@tobiasheine


blundell:
@JozefCeluch can we get screenshots on any PR's that affect UI?

JozefCeluch:
@blundell updated

PR #38, opened 2016-05-13
React/PT-70/tweet detail

danybony:
Adds a Tweet detail screen.
This screen doesn't show more information (for now) compared to a single tweet item in the list but it's good to see how to implement master-detail navigation.
It's possible to access the detail view by selecting a tweet from the list (but then it's not loading the tweet, due to missing OAuth valid token) or directly from the debug screen (and then passing a hardcoded tweet id).
Also, implementing the Android hardware back navigation (no conflicts with iOS, given in that case all the methods are no-op)

| iOS  | Android |
| --- | --- |
| screen shot 2016-05-13 at 17 23 40 | screen shot 2016-05-13 at 17 23 17 |

### DIFF
--- a/ReactNative/ReactTwitter/core/views/debug-screen.js
+++ b/ReactNative/ReactTwitter/core/views/debug-screen.js
@@ -25,6 +25,10 @@ var DebugScreenView = React.createClass({
           style={styles.button}
           styleDisabled={styles.button_disabled}
           onPress={this.pushTweetsList}>Tweets List</Button>
+        <Button
+          style={styles.button}
+          styleDisabled={styles.button_disabled}
+          onPress={this.pushTweetView}>Tweet Detail View</Button>
       </View>
     )
   },
@@ -35,6 +39,10 @@ var DebugScreenView = React.createClass({
 
   pushTweetsList () {
     this.props.navigator.push({id: 'tweets-list-identifier'})
+  },
+
+  pushTweetView () {
+    this.props.navigator.push({id: 'tweet-view-identifier'})
   }
 })
 
--- a/ReactNative/ReactTwitter/core/views/debug-screen.js
+++ b/ReactNative/ReactTwitter/core/views/debug-screen.js
@@ -42,7 +42,10 @@ var DebugScreenView = React.createClass({
   },
 
   pushTweetView () {
-    this.props.navigator.push({id: 'tweet-view-identifier'})
+    this.props.navigator.push({
+      id: 'tweet-view-identifier',
+      tweetId: '210462857140252672'
+    })
   }
 })
 
--- a/ReactNative/ReactTwitter/core/views/mixins/android-back-navigation.js
+++ b/ReactNative/ReactTwitter/core/views/mixins/android-back-navigation.js
@@ -1,0 +1,25 @@
+import React from 'react'
+import { BackAndroid, Navigator } from 'react-native'
+
+/*eslint-disable no-unused-vars*/
+var AndroidBackNavigationMixin = {
+  propTypes: {
+    navigator: React.PropTypes.instanceOf(Navigator).isRequired
+  },
+
+  componentDidMount () {
+    BackAndroid.addEventListener('hardwareBackPress', () => {
+      if (this.props.navigator.getCurrentRoutes().length === 1) {
+        return false
+      }
+      this.props.navigator.pop()
+      return true
+    })
+  },
+
+  componentWillUnmount: function () {
+    BackAndroid.removeEventListener('hardwareBackPress')
+  }
+}
+
+module.exports = AndroidBackNavigationMixin
--- a/ReactNative/ReactTwitter/core/views/mixins/android-back-navigation.js
+++ b/ReactNative/ReactTwitter/core/views/mixins/android-back-navigation.js
@@ -8,17 +8,20 @@ var AndroidBackNavigationMixin = {
   },
 
   componentDidMount () {
-    BackAndroid.addEventListener('hardwareBackPress', () => {
-      if (this.props.navigator.getCurrentRoutes().length === 1) {
-        return false
-      }
-      this.props.navigator.pop()
-      return true
-    })
+    BackAndroid.addEventListener('hardwareBackPress', this._handleBackAndroid)
   },
 
   componentWillUnmount: function () {
-    BackAndroid.removeEventListener('hardwareBackPress')
+    BackAndroid.removeEventListener('hardwareBackPress', this._handleBackAndroid)
+  },
+
+  _handleBackAndroid () {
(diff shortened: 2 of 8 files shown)


ReactNative/ReactTwitter/core/views/debug-screen.js
```diff
@@ -35,6 +39,13 @@ var DebugScreenView = React.createClass({
 
   pushTweetsList () {
     this.props.navigator.push({id: 'tweets-list-identifier'})
+  },
+
+  pushTweetView () {
+    this.props.navigator.push({
+      id: 'tweet-view-identifier',
+      tweetId: '210462857140252672'
```
danybony: This is how we pass extra params to the navigator route, which can be read from the receiving component accessing its props.

ReactNative/ReactTwitter/core/views/tweetView.js
```diff
@@ -0,0 +1,134 @@
...
+      return (<View />)
+    }
+
+    var formattedTime = this._formatTime(Date.parse(this.state.tweet.created_at))
+    return (
+      <View style={styles.mainContainer}>
+        <View style={styles.header}>
+          <Image style={styles.tweet_avatar} source={{ uri: this.state.tweet.user.profile_image_url }} />
+          <View style={styles.textContainer}>
+            <View>
```
danybony: This is the horrible way I found to have the two following `<Text />` center vertically together, and aligned to the left.
If someone can find a better way...

ReactNative/ReactTwitter/core/views/tweetView.js
```diff
@@ -0,0 +1,134 @@
...
+    )
+  },
+
+  _formatTime (time) {
+    var dateFormat = require('dateformat')
+    return dateFormat(time, 'mmmm dS, yyyy, h:MM:ss TT')
+  },
+
+  _tweetSelected (tweetId) {
+    // TODO: navigate to tweet details screen
```
eduardb: Aren't we already in the tweet details screen? 👅

ReactNative/ReactTwitter/core/views/tweetView.js
```diff
@@ -0,0 +1,124 @@
...
+        'Authorization': 'OAuth oauth_consumer_key="aqPSTs1FT2ndP7qXi247BtbXd", oauth_nonce="9e48307c7ed7184557a3d88f6331f00f", oauth_signature="Jr9YqRlJ%2BNH%2BBgLC9cI%2F7LD06qg%3D", oauth_signature_method="HMAC-SHA1", oauth_timestamp="1463143567", oauth_token="730013266697654273-RDssoTCOdHNQtFA9k87OSijeZHcF4SU", oauth_version="1.0"'
+      }
+    })
+      .then((response) => response.json())
+      .then((rjson) => {
+        this.setState({
+          tweet: rjson
+        })
+      })
+      .done()
```
frapontillo: There's no need for this if you don't pass anything into it.

ReactNative/ReactTwitter/core/views/tweetView.js
```diff
@@ -0,0 +1,124 @@
...
+          </View>
+        </View>
+        <Text style={styles.tweet_time}>{formattedTime}</Text>
+        <Text style={styles.tweet_text}>{this.state.tweet.text}</Text>
+      </View>
+    )
+  },
+
+  _formatTime (time) {
+    var dateFormat = require('dateformat')
```
frapontillo: Require this on top of the file. It makes all dependencies explicit (I think you can also use `import`).

ReactNative/ReactTwitter/core/views/tweetsList.js
```diff
@@ -35,7 +48,7 @@ class TweetsList extends Component {
         })
       })
       .done()
```
frapontillo: As before.

frapontillo:
A few comments left.
(loving react native, uh?)

ReactNative/ReactTwitter/core/views/tweetView.js
```diff
@@ -0,0 +1,124 @@
...
+  },
+
+  getInitialState () {
+    return {
+      tweet: {}
+    }
+  },
+
+  componentDidMount () {
+    BackAndroid.addEventListener('hardwareBackPress', () => {
```
gbasile: we probably need this behaviour on different Screen, Is it worth to create a super-class? Or some kind of mixin?

ReactNative/ReactTwitter/core/views/tweetView.js
```diff
@@ -0,0 +1,124 @@
...
+      .then((rjson) => {
+        this.setState({
+          tweet: rjson
+        })
+      })
+  },
+
+  render () {
+    var tweetx = this.state.tweet
+    console.log('render')
```
gbasile: Going to remove these logs

ReactNative/ReactTwitter/core/views/tweetView.js
```diff
@@ -0,0 +1,124 @@
...
+            </View>
+          </View>
+        </View>
+        <Text style={styles.tweet_time}>{formattedTime}</Text>
+        <Text style={styles.tweet_text}>{this.state.tweet.text}</Text>
+      </View>
+    )
+  },
+
+  _formatTime (time) {
```
gbasile: Probably this formatter should be implemented somewhere else if we need it in other screens

ReactNative/ReactTwitter/core/views/tweetsList.js
```diff
@@ -1,20 +1,33 @@
...
     var ds = new ListView.DataSource({rowHasChanged: (r1, r2) => r1 !== r2})
-    this.state = {
+    return {
       dataSource: ds.cloneWithRows([])
     }
-  }
+  },
 
   componentDidMount () {
+    BackAndroid.addEventListener('hardwareBackPress', () => {
```
gbasile: That's the reason why we should implement this somewhere else, we already have this code duplicated in 2 screens

ReactNative/ReactTwitter/core/views/mixins/android-back-navigation.js
```diff
@@ -0,0 +1,25 @@
...
+      if (this.props.navigator.getCurrentRoutes().length === 1) {
+        return false
+      }
+      this.props.navigator.pop()
+      return true
+    })
+  },
+
+  componentWillUnmount: function () {
+    BackAndroid.removeEventListener('hardwareBackPress')
```
eduardb: Shouldn't you provide also a reference to the `handler` that you added previously?

ReactNative/ReactTwitter/core/views/tweetView.js
danybony: Ah yeah, my bad, I totally forgot about those :(

PR #39, opened 2016-05-17
React/pt 68/login screen flow

eduardb:
TL;DR Proper screen  flow on app startup: splash screen -> user logged in ? tweets timeline : login screen

We decided to have a splash screen as the entry point in the app because we have to load asynchronously the stored user credentials, if any.  If there are no stored credentials, we go to the login screen, else to the user tweets timeline (which is broken atm until we introduce the oauth stuff there as well).

We also decided to move the components related to debug in a `debug` subfolder.

The debug screen is inaccessible atm, as soon as we introduce a toolbar / nav bar, we will add a menu item to access it again :D

| iOS  | Android |
| --- | --- |
| screenshot-ios | screenshot-android | 

Paired with @gbasile


ReactNative/ReactTwitter/core/views/splash-screen.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+  },
+
+  componentDidMount () {
+    let authService = this.state.authenticationService
+
+    authService.loadDataFromDisk()
+      .then(() => {
+        return new Promise((resolve, reject) => {
+          setTimeout(resolve, 1100)
```
takecare: why 1100?

eduardb: it was carefully chosen for a good UX impact 👅 
I will extract it to a constant

ReactNative/ReactTwitter/core/views/login-screen.js
```diff
@@ -0,0 +1,106 @@
...
+        .catch(console.warn)
+  },
+
+  _browserAuthenticationWithToken (oauthToken) {
+    this.state.facade.listenForDeepLinking().then((uri) => {
+      let parsedURI = OauthHelper.getOauthTokenAndVerifierFromURLCallback(uri)
+      this.state.facade.stopListeningForDeepLinking()
+
+      if (parsedURI.oauth_token == null || parsedURI.oauth_token.length === 0) {
+        // user canceled login
```
xrigau: this can be replaced with `if (userCancelledLogin())` 👅

ReactNative/ReactTwitter/core/views/login-screen.js
```diff
@@ -0,0 +1,106 @@
...
+            } else {
+              console.log('Something went wrong :o')
+            }
+          })
+        })
+    })
+    Linking.openURL('https://api.twitter.com/oauth/authenticate?oauth_token=' + oauthToken)
+  },
+
+  _pushTweetsList () {
```
xrigau: is the leading `_` a convention?

eduardb: Yes, it is :P

ReactNative/ReactTwitter/core/views/login-screen.js
eduardb: I will take into consideration your suggestion for a future PR

ReactNative/ReactTwitter/core/views/login-screen.js
xrigau: 👍 same with the space before the `(` 😷

eduardb: The linter will complain if we don't put the space. Probably that we could alter the rules, but it's a bit too late for that probably.

gbasile: We are using the default preset for the linter, I would imagine this is pretty standard

`{
    "extends": ["standard", "plugin:react/recommended"],
    "plugins": [
        "standard",
        "react"
    ]
}`

PR #41, opened 2016-05-17
React/pt 68/twitter login

gbasile:
TL;DR;

Implement the twitter login with oauth 1.0 and related views. 

All the content of this PR has been already reviewed, this is here just to have another look to the entire list of changes


ReactNative/ReactTwitter/core/views/splash-screen.js
```diff
@@ -0,0 +1,71 @@
...
+          setTimeout(resolve, SPLASH_DURATION_MS)
+        })
+      })
+      .then(() => {
+        if (authService.isUserLoggedIn()) {
+          this._pushTweetsList()
+        } else {
+          this._pushLogin()
+        }
+      })
```
gbasile: I know it's never the case, but should we also put here a catch statement for possible errors?

eduardb: Hmm, and maybe redirecting the user to the login screen in case of an error?

PR #46, opened 2016-05-18
React/pt 70/tweets list oauth integration

eduardb:
TL;DR Use real data for the tweets list and tweet details

Previously, the authorization header for the aforementioned calls was hardcoded. So now we use the `OauthHelper` to compute the real auhorization header. To achieve this, we defined methods in the `TwitterRequestsService` to help us for that, and added support for query parameters in `OauthHelper`. We also inject the `TwitterRequestsService` across the different views so we don't have to instantiate each time a new service, and we encapsulated the `AuthenticationService` logic in the service. Plus misc refactors to improve quality.

No actual screenshots, as we only touched the functionality, not the UI.

Paired with @gbasile

### DIFF
--- a/ReactNative/ReactTwitter/core/views/login-screen.js
+++ b/ReactNative/ReactTwitter/core/views/login-screen.js
@@ -57,7 +57,7 @@ var LoginScreenView = React.createClass({
       let parsedURI = OauthHelper.getOauthTokenAndVerifierFromURLCallback(uri)
       this.state.facade.stopListeningForDeepLinking()
 
-      if (parsedURI.oauth_token == null || parsedURI.oauth_token.length === 0) {
+      if (this._userCancelledLogin(parsedURI)) {
         // user canceled login
         return
       }
@@ -75,6 +75,10 @@ var LoginScreenView = React.createClass({
         })
     })
     Linking.openURL('https://api.twitter.com/oauth/authenticate?oauth_token=' + oauthToken)
+  },
+
+  _userCancelledLogin (parsedURI) {
+    return parsedURI.oauth_token == null || parsedURI.oauth_token.length === 0
   },
 
   _pushTweetsList () {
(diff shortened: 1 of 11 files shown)


ReactNative/ReactTwitter/core/views/tweetsList.js
```diff
@@ -23,23 +25,14 @@ var TweetsList = React.createClass({
...
-      this.setState({
-        dataSource: this.state.dataSource.cloneWithRows(rjson)
+    this.props.twitterService.getHomeTimeline()
+      .then((rjson) => {
+        this.setState({
+          dataSource: this.state.dataSource.cloneWithRows(rjson)
+        })
       })
-    })
+      .catch((err) => { console.warn(err) })
```
xrigau: 👍

dominicfreeston:
I've looked at this multiple times now but I lack too much context to provide any useful feedback, but 👍 from me FWIW.

ReactNative/ReactTwitter/core/views/login-screen.js
```diff
@@ -11,18 +11,16 @@ var Button = require('react-native-button')
 var TwitterRequestsService = require('../service/twitter-requests-service.js')
 var DeepLinkingFacade = require('../service/deep-linking-facade')
 var OauthHelper = require('../service/oauth-helper.js')
-var AuthenticationService = require('../service/authentication-service.js')
 
 var LoginScreenView = React.createClass({
   propTypes: {
-    navigator: React.PropTypes.instanceOf(Navigator).isRequired
+    navigator: React.PropTypes.instanceOf(Navigator).isRequired,
+    twitterService: React.PropTypes.instanceOf(TwitterRequestsService).isRequired
```
dominicfreeston: [Question] Does `isRequired` set a constraint, or does it actually create the instance here?

eduardb: It's just a constraint for validation, the instance is sent using the component's `props` from here: https://github.com/novoda/spikes/blob/react/PT-70/tweets-list-oauth-integration/ReactNative/ReactTwitter/core/views/main-navigator.js#L37

xrigau:
:shipit:

PR #48, opened 2016-05-19
Create channels!

yvettecook:
User can now create channels from the app.

create-channel

### DIFF
--- a/Firebase/ios/Bonfire/Mobile/Utilities/Observable+Extensions.swift
+++ b/Firebase/ios/Bonfire/Mobile/Utilities/Observable+Extensions.swift
@@ -1,0 +1,6 @@
+import Foundation
+import RxSwift
+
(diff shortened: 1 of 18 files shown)


Firebase/ios/Bonfire/Mobile/Utilities/Observable+Extensions.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+import RxSwift
+
+func mergeToArray<T>(observables: [Observable<T>]) -> Observable<[T]> {
```
dominicfreeston: I know I was pretty responsible for this name, but maybe we can be more accurate / descriptive?

`convertToObservableArray` maybe?

`convertArrayOfObservablesToObservableArray` too much?

yvettecook: Agreed - will do in next PR

PR #50, opened 2016-05-20
Update channel service to reflect the new channel structure

JozefCeluch:
Updates the FIrebaseChannelService to reflect the updated structure of channels and their indexes.

### DIFF
--- a/Firebase/android/app/src/main/java/com/novoda/bonfire/channel/service/FirebaseChannelService.java
+++ b/Firebase/android/app/src/main/java/com/novoda/bonfire/channel/service/FirebaseChannelService.java
@@ -45,25 +45,8 @@ public class FirebaseChannelService implements ChannelService {
         return Observable.create(new Observable.OnSubscribe<List<String>>() {
             @Override
             public void call(final Subscriber<? super List<String>> subscriber) {
-                final DatabaseReference privateChannelsOfUserDb = FirebaseChannelService.this.privateChannelsDB.child(user.getId());
-                final ValueEventListener eventListener = privateChannelsOfUserDb.addValueEventListener(new ValueEventListener() {
-                    @Override
-                    public void onDataChange(DataSnapshot dataSnapshot) {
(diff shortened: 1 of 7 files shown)


Firebase/android/app/src/main/java/com/novoda/bonfire/channel/service/FirebaseChannelService.java
```diff
@@ -97,15 +95,107 @@ public void call() {
...
+                return Observable.create(new Observable.OnSubscribe<ChannelInfo>() {
+                    @Override
+                    public void call(final Subscriber<? super ChannelInfo> subscriber) {
+                        channelsDB.child(channelPath).addListenerForSingleValueEvent(new ValueEventListener() {
+                            @Override
+                            public void onDataChange(DataSnapshot dataSnapshot) {
+                                if (dataSnapshot.hasChildren()) {
+                                    subscriber.onNext(toChannelInfo(dataSnapshot));
+                                }
+                                subscriber.onCompleted();
```
takecare: so deep! can't we extract some of this stuff?

Firebase/android/app/src/main/java/com/novoda/bonfire/channel/service/FirebaseChannelService.java
```diff
@@ -1,11 +1,14 @@
 package com.novoda.bonfire.channel.service;
 
+import android.support.annotation.NonNull;
```
takecare: sure to add this dependency here?

JozefCeluch: this class is in the app module

JozefCeluch:
@takecare your comments have been addressed

joetimmins:
Much more readable now without the nested Observables. I think it's always going to be a lot of Rx processing stuff going on here, because the data structure itself is very convoluted.

👍

PR #51, opened 2016-05-20
Add pull-to-refresh mechanism for tweets list

eduardb:
Basically, added a `RefreshControl` as a `refreshControl` property of the `ListView`

BEFORE: 

| iOS | Android |
| --- | --- |
| ios-before | ezgif com-video-to-gif |

AFTER: 

| iOS | Android |
| --- | --- |
| ios-after | ezgif com-video-to-gif 1 |

Paired with no one


xrigau:
wow react native seems so EASY! :trollface:

PR #52, opened 2016-05-23
React/pt 110/ui improvement

gbasile:
All these changes are UI related to makes the interface looks a little bit better.

may-23-2016 12-04-08


ReactNative/ReactTwitter/core/views/common/avatar-view.js
```diff
@@ -0,0 +1,34 @@
...
+    })
+
+    return <Image
+      source= {source}
+      style = {style.image}
+      resizeMode='contain'/>
+  }
+
+  _getHighResolutionImageURL (originalURL) {
+    return originalURL.replace('_normal', '_bigger')
```
danybony: Are these two always available for twitter avatar pictures?

gbasile: Yes, this is documented here https://dev.twitter.com/overview/general/user-profile-images-and-banners

ReactNative/ReactTwitter/core/views/common/data-format.js
```diff
@@ -1,9 +1,13 @@
 var dateFormat = require('dateformat')
-
+var tinyHumanTime = require('tiny-human-time')
 class NovodaDataFormat {
   static longTime (time) {
     return dateFormat(time, 'mmmm dS, yyyy, h:MM:ss TT')
   }
+
+  static elapsedTime (time) {
+    return tinyHumanTime(new Date(), time) + ' ago'
```
danybony: 👍

ReactNative/ReactTwitter/core/views/common/avatar-view.js
danybony: Cool! 👍

ReactNative/ReactTwitter/core/views/splash-screen.js
```diff
@@ -49,7 +50,7 @@ var SplashScreenView = React.createClass({
   render () {
     return (
       <View style={styles.container}>
-        <Text style={styles.title}>SplashScreen</Text>
+        <Image style={styles.logo} source={require('./assets/twitter-logo.png')} />
```
danybony: Loading it like that, is it going to work for both platforms? (I guess so)
It would be good to see how to include assets for different screen densities. In android we have this problem, how is it in iOS? (different screens?)

ReactNative/ReactTwitter/core/views/tweetsListItem.js
```diff
@@ -24,16 +24,18 @@ var TweetsListItem = React.createClass({
...
   render () {
-    var formattedTime = novodaDataFormat.longTime(this.props.time)
+    var formattedTime = novodaDataFormat.elapsedTime(this.props.time)
     return (
       <TouchableHighlight onPress={() => this._tweetSelected(this.props.id)}
         underlayColor='#dddddd'>
          <View>
           <View style={styles.rowContainer}>
-            <Image style={styles.tweet_avatar} source={{ uri: this.props.author_avatar }} />
+            <AvatarView uri={this.props.author_avatar} size={60} />
```
danybony: I wonder if we can take the size from a style...

ReactNative/ReactTwitter/core/views/splash-screen.js
gbasile: We can use images with differents densities using the suffix: @2x and @3x. 

From the documentation: 

> Packager will bundle and serve the image corresponding to device's screen density, e.g. on iPhone 5s check@2x.png will be used, on Nexus 5 – check@3x.png. If there is no image matching the screen density, the closest best option will be selected.

ReactNative/ReactTwitter/core/views/tweetsListItem.js
gbasile: AvatarView is a custom component that I have created. I decided to pass the size as a props so that internally it uses it for both the width and the height.

alexstyl:
👍  LGTM

@danybony are you happy to merge this?

takecare:
where's the android sshot?

PR #56, opened 2016-05-24
React/flow configuration

gbasile:
Flow is a static type checker for JavaScript. It can be used to catch common bugs in JavaScript programs before they run

- Configured flow to work correctly in the project
- Updated Eslint to support flow syntax
- Annotated files with `// @flow` to enable static type checking
- Fixed all the errors


danybony:
Well done, especially fixing all the reported issues!
LGTM 👍 
(the `// @flow` comment to enable the analysis is so ugly though...)

PR #59, opened 2016-05-25
Push Notification iOS

gbasile:
Nothing really relevant to comment about this PR, I just need to bridge some event from the app to the react-native component. When a notification is received, an alert pop up. 

pushnotification1
pushnotification2

### DIFF
--- a/ReactNative/ReactTwitter/core/views/debug/push-notifications-screen.js
+++ b/ReactNative/ReactTwitter/core/views/debug/push-notifications-screen.js
@@ -35,8 +35,9 @@ var PushNotificationsScreenView = React.createClass({
         <Button
           style={styles.button}
           styleDisabled={styles.button_disabled}
-          onPress={this._buttonClicked}> Send push notification</Button>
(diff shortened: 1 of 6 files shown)


eduardb:
LGTM (I have no idea what you did)

ReactNative/ReactTwitter/core/views/debug/push-notifications-screen.js
```diff
@@ -35,8 +35,9 @@ var PushNotificationsScreenView = React.createClass({
         <Button
           style={styles.button}
           styleDisabled={styles.button_disabled}
-          onPress={this._buttonClicked}> Send push notification</Button>
+          onPress={this._buttonClicked}> Send Android push notification</Button>
```
takecare: why's this "android only"?

gbasile: This button send a push notification to test the functionality. 
This feature has been de-scoped and it will be probably deleted from the debug screen has we don't want to deploy private keys nor certificates.

PR #62, opened 2016-05-26
Added Debug Screen for Native Component Screen

gbasile:
Added the debug screen for the native component, it already implement a basic ListView
Debug Screen is now a ListView easily configurable with new screens
Refactor navigator so we get rid of magic string constants
Each screen now implement static methods for the navigatorID and 

screen shot 2016-05-26 at 16 38 33

### DIFF
--- a/ReactNative/ReactTwitter/core/views/debug/debug-screen.js
+++ b/ReactNative/ReactTwitter/core/views/debug/debug-screen.js
@@ -13,12 +13,14 @@ var DeepLinkingView = require('./deep-linking.js')
 var PushNotificationsScreenView = require('./push-notifications-screen.js')
 var OauthView = require('./oauth-screen.js')
 var DebugAnimationScene = require('./debug-animation-scene.js')
+var DebugNativeComponentScene = require('./debug-native-component-scene.js')
 
 const scenes = [
   DeepLinkingView,
   PushNotificationsScreenView,
   OauthView,
   DebugAnimationScene,
+  DebugNativeComponentScene
 ]
 
 var DebugScreenView = React.createClass({
--- a/ReactNative/ReactTwitter/core/views/debug/debug-screen.js
+++ b/ReactNative/ReactTwitter/core/views/debug/debug-screen.js
@@ -37,7 +37,8 @@ var DebugScreenView = React.createClass({
         dataSource={this.state.dataSource}
         enableEmptySections={true}
         renderRow={this._renderRow}
-        style={styles.container}
+        style={styles.list}
+        contentContainerStyle={styles.container}
       />
     )
   },
@@ -55,13 +56,16 @@ var DebugScreenView = React.createClass({
 })
 
 const styles = StyleSheet.create({
-  container: {
+  list: {
     backgroundColor: '#F5FCFF'
+  },
+  container: {
+    alignItems: 'flex-start'
   },
   button: {
     fontSize: 20,
     color: 'black',
-    margin: 6
+    margin: 8
   },
   button_disabled: {
     fontSize: 20,
--- a/ReactNative/ReactTwitter/core/views/debug/debug-screen.js
+++ b/ReactNative/ReactTwitter/core/views/debug/debug-screen.js
@@ -3,8 +3,8 @@
 import React from 'react'
 import {
   StyleSheet,
-  View,
-  Navigator
+  Navigator,
+  ListView
 } from 'react-native'
 
 var Button = require('react-native-button')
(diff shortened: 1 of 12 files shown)


eduardb:
> Refactor navigator so we get rid of magic string constants

But some magic string constants are still used to navigate to screens. For instance, in `TweetsListItem. _tweetSelected()`. Is it possible to get rid of them as well?

gbasile:
@eduardb good catch. Issued

ReactNative/ReactTwitter/core/views/debug/debug-screen.js
```diff
@@ -3,59 +3,88 @@
...
   propTypes: {
     navigator: React.PropTypes.instanceOf(Navigator).isRequired
   },
 
+  getInitialState: function () {
+    var ds = new ListView.DataSource({rowHasChanged: (r1, r2) => r1 !== r2})
+    let rows = scenes.map((scene) => {
+      return {
+        name: scene.navigatorTitle(),
+        sceneID: scene.navigatorID()
```
danybony: Is there a way to enforce that all the scenes have these two methods implemented?
As it is now we have a crash at runtime only, isn't it?

gbasile: AFAIK there is no support for interfaces/abstract classes. 
Please @frapontillo enlighten us

ReactNative/ReactTwitter/core/views/main-navigator.js
```diff
@@ -64,28 +68,15 @@ var NavigationBarRouteMapper = {
...
-    }
+    let title = ''
+    scenes.forEach((scene) => {
+      if (scene.navigatorID() === screenId) {
+        title = scene.navigatorTitle()
+        return
+      }
+    })
+
+    return title
```
danybony: I would Throw an exception here: fail fast

ReactNative/ReactTwitter/core/views/debug/debug-screen.js
danybony: Might be the holy light of @frapontillo be within us! [](http://i.giphy.com/6gbpsa2yV4tMY.gif)

danybony:
Very good update of the navigation 💯 
Can we have a screenshot of the new debug screen?

gbasile:
I can put a screenshot of the new debug screen, but it looks exactly the same :)

ReactNative/ReactTwitter/core/views/debug/debug-screen.js
frapontillo: I don't think it's possible yet.
If you need to be this powerful I strongly advise you to move to TypeScript, which [has interfaces](https://www.typescriptlang.org/docs/handbook/interfaces.html) and is a language on its own.

To circumvent the limitation, you can create your own `StatefulClassCreator` with a `.createClass` method that checks that the passed object has the `getInitialState` property and, if so, delegates it to `React.createClass`.

gbasile: and then we have to remember to instantiate the class correctly 😂

ReactNative/ReactTwitter/core/views/main-navigator.js
```diff
@@ -64,28 +68,15 @@ var NavigationBarRouteMapper = {
...
-      case debugAnimationSceneID:
-        return 'Debug Animation'
-      default:
-        return ''
-    }
+    let title = ''
+    scenes.forEach((scene) => {
+      if (scene.navigatorID() === screenId) {
+        title = scene.navigatorTitle()
+        return
```
frapontillo: This does not work as you'd expect it would: all you get from this `return` is to exit the current iteration, which you still get if this is the last line 😄 

If you want to break of the loop, you need to use a `for ... in` and `return` from there.

ReactNative/ReactTwitter/core/views/main-navigator.js
```diff
@@ -100,31 +91,33 @@ var MainNavigator = React.createClass({
...
         navigationBar={navBar}
-        initialRoute={{id: splashScreenID}}
+        initialRoute={{id: SplashScreenView.navigatorID()}}
         renderScene={this.navigatorRenderScene}
       />)
   },
 
   navigatorRenderScene (route: any, navigator: Navigator) {
     switch (route.id) {
-      case splashScreenID:
```
frapontillo: I'd actually rewrite this into an object with key-values matching the case and string you need to return, so it's simpler to read and accessing the right template is just a matter of accessing a dictionary value.

ReactNative/ReactTwitter/core/views/main-navigator.js
danybony: Ah you are right, this is just returning for the function
```
(scene) => {
       if (scene.navigatorID() === screenId) {
         title = scene.navigatorTitle()
         return
}
```
So this is wrong

ReactNative/ReactTwitter/core/views/main-navigator.js
danybony: Well, throwing an exception only if it's not found

ReactNative/ReactTwitter/core/views/main-navigator.js
gbasile: I feel so n00b....

giphy

ReactNative/ReactTwitter/core/views/debug/debug-screen.js
frapontillo: Of course you do! 💃

ReactNative/ReactTwitter/core/views/main-navigator.js
gbasile: Fixed

ReactNative/ReactTwitter/core/views/main-navigator.js
```diff
@@ -68,15 +68,13 @@ var NavigationBarRouteMapper = {
   },
 
   _getTitleByScreenId (screenId) {
-    let title = ''
-    scenes.forEach((scene) => {
+    for (let scene of scenes) {
```
frapontillo: 👍

PR #64, opened 2016-05-26
Added android package name to dynamic link,

yvettecook:
... and some imports were added that were required to prevent build errors.

Basically, dynamic links are now cross platform 🎉

### DIFF
--- a/Firebase/ios/Bonfire.xcodeproj/project.pbxproj
+++ b/Firebase/ios/Bonfire.xcodeproj/project.pbxproj
@@ -711,13 +711,10 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_ENTITLEMENTS = Bonfire/Bonfire.entitlements;
-				CODE_SIGN_IDENTITY = "iPhone Developer";
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				INFOPLIST_FILE = Bonfire/Info.plist;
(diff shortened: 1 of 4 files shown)


Firebase/ios/Bonfire/Mobile/FirebaseRemoteConfig.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 import Firebase
+import FirebaseRemoteConfig
```
dominicfreeston: Weird that I'm not getting build errors - I've started doing just `import Firebase` and it seems to include everything, as the docs claim it should

Firebase/ios/Bonfire.xcodeproj/project.pbxproj
```diff
@@ -711,10 +711,13 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_ENTITLEMENTS = Bonfire/Bonfire.entitlements;
+				CODE_SIGN_IDENTITY = "iPhone Developer";
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
```
dominicfreeston: Why do we now need this explicitly defined?

dominicfreeston:
Got Xcode-related questions, but otherwise 👍

Firebase/ios/Bonfire/Mobile/FirebaseRemoteConfig.swift
yvettecook: It's a mystery - without it I cannot build

PR #66, opened 2016-05-27
React/PT-131/Android accessibility

danybony:
This PR investigates the following aspects of accessibility:

- Add a custom text read from TalkBack instead of the default one
- Ignore certain elements of a group when navigating (for example in a list)

Also uses the Autolink library to automatically highlight links inside a text and handle click.
This is not working with accessibility though (it's not possible to select the link inside the text when navigating with talkback on). An issue will be opened to the library repo.

image

Paired with ☕

### DIFF
--- a/ReactNative/ReactTwitter/core/views/common/avatar-view.js
+++ b/ReactNative/ReactTwitter/core/views/common/avatar-view.js
@@ -25,6 +25,7 @@ class AvatarView extends Image {
     return <Image
       source= {source}
       style = {style.image}
+      importantForAccessibility="no"
       resizeMode='contain'/>
   }
 
--- a/ReactNative/ReactTwitter/core/views/common/avatar-view.js
+++ b/ReactNative/ReactTwitter/core/views/common/avatar-view.js
@@ -25,7 +25,7 @@ class AvatarView extends Image {
     return <Image
       source= {source}
       style = {style.image}
-      importantForAccessibility="no"
+      importantForAccessibility='no'
       resizeMode='contain'/>
   }
 
--- a/ReactNative/ReactTwitter/core/views/tweetView.js
+++ b/ReactNative/ReactTwitter/core/views/tweetView.js
@@ -8,9 +8,10 @@ import {
   Navigator
 } from 'react-native'
 import AndroidBackNavigationMixin from './mixins/android-back-navigation'
-var novodaDataFormat = require('./common/data-format.js')
-var TwitterRequestsService = require('../service/twitter-requests-service.js')
-var AvatarView = require('./common/avatar-view.js')
+const novodaDataFormat = require('./common/data-format.js')
+const TwitterRequestsService = require('../service/twitter-requests-service.js')
+const AvatarView = require('./common/avatar-view.js')
+const AutoLink = require('react-native-autolink')
 
 var TweetsView = React.createClass({
   mixins: [AndroidBackNavigationMixin],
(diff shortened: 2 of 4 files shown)


ReactNative/ReactTwitter/core/views/common/avatar-view.js
```diff
@@ -25,6 +25,7 @@ class AvatarView extends Image {
     return <Image
       source= {source}
       style = {style.image}
+      importantForAccessibility='no'
```
danybony: The image is ignored when TalkBack is on

ReactNative/ReactTwitter/core/views/tweetsListItem.js
```diff
@@ -36,10 +37,10 @@ var TweetsListItem = React.createClass({
             <View style={styles.textContainer}>
               <View style={styles.userInfoContainer}>
                 <Text style={styles.tweet_author}>{this.props.author_name}</Text>
-                <Text style={styles.tweet_author_handle}>@{this.props.author_handle}</Text>
+                <Text importantForAccessibility='no' style={styles.tweet_author_handle}>@{this.props.author_handle}</Text>
```
danybony: Ignore the user handle when navigating between the list items

ReactNative/ReactTwitter/core/views/tweetsListItem.js
```diff
@@ -36,10 +37,10 @@ var TweetsListItem = React.createClass({
...
               </View>
               <Text style={styles.tweet_time}>{formattedTime}</Text>
-              <Text style={styles.tweet_text}>{this.props.text}</Text>
+              <Text style={styles.tweet_text} accessibilityLabel={textWithoutUrls}>{this.props.text}</Text>
```
danybony: Replace the read text (not the displayed one) with one without urls

danybony:
It would be good to have a feedback from @ataulm as well, if we could easily test something else

ReactNative/ReactTwitter/core/views/tweetsListItem.js
ataulm: do these work just by mapping to the xml attrs?

ReactNative/ReactTwitter/core/views/tweetView.js
```diff
@@ -8,9 +8,10 @@ import {
   Navigator
 } from 'react-native'
 import AndroidBackNavigationMixin from './mixins/android-back-navigation'
-var novodaDataFormat = require('./common/data-format.js')
-var TwitterRequestsService = require('../service/twitter-requests-service.js')
-var AvatarView = require('./common/avatar-view.js')
+const novodaDataFormat = require('./common/data-format.js')
```
ataulm: shoudl this be `NovodaDataFormat`?

ataulm:
:+1:

ataulm:
Is it easy to determine if talkback is enabled and present a different UI instead?

ReactNative/ReactTwitter/core/views/tweetsListItem.js
danybony: yes exactly, these attributes are passes as properties for the component constructor, in this case `Text`

ReactNative/ReactTwitter/core/views/tweetView.js
danybony: 👍  good catch

danybony:
@ataulm I haven't find that.
It should be possible creating a specific Android component, but haven't had time for that

PR #68, opened 2016-06-01
Invite to the app

Dorvaryn:
This PR adds the ability to invite someone to the App using Dynamic links.

| Before | After |
| ------ | ----- |
| No invite system | Invite

The Invites work across ios and Android


Firebase/android/app/src/main/java/com/novoda/bonfire/link/FirebaseDynamicLinkFactory.java
```diff
@@ -0,0 +1,42 @@
...
+                .appendQueryParameter("apn", androidPackageName)
+                .build();
+        return URI.create(uri.toString());
+    }
+
+    private Uri welcomeDeepLinkFromUser(User user) {
+        return Uri.parse(deepLinkBaseUrl)
+                .buildUpon()
+                .appendPath("welcome")
+                .appendQueryParameter("sender", user.getName())
```
JozefCeluch: The query parameter is used to retrieve the name from the intent later. Should probably be a constant somewhere?

JozefCeluch:
👍  LGTM but I might be biased 😄 
@joetimmins would you have time to have a look?

Firebase/android/app/src/main/java/com/novoda/bonfire/BaseActivity.java
```diff
@@ -6,7 +6,7 @@
 
 import com.novoda.notils.logger.simple.Log;
 
-public class BaseActivity extends AppCompatActivity {
+public abstract class BaseActivity extends AppCompatActivity {
```
dominicfreeston: so abstract

Firebase/android/app/src/main/java/com/novoda/bonfire/link/FirebaseDynamicLinkFactory.java
```diff
@@ -0,0 +1,43 @@
+package com.novoda.bonfire.link;
+
+import android.net.Uri;
+
+import com.novoda.bonfire.user.data.model.User;
+
+import java.net.URI;
+
+public class FirebaseDynamicLinkFactory implements LinkFactory {
```
dominicfreeston: We called the interface `DynamicLinkFactory`. I mean, it doesn't matter that much, but just thought I'd point it out.

dominicfreeston:
LGTM FWIW AORA(and other relevant acronyms)

joetimmins:
Hadn't come across AORA before 😂

Firebase/android/app/src/main/AndroidManifest.xml
```diff
@@ -32,6 +31,22 @@
...
+        <category android:name="android.intent.category.DEFAULT" />
+        <category android:name="android.intent.category.BROWSABLE" />
+        <data
+          android:host="novoda.com"
+          android:pathPrefix="/bonfire/welcome"
+          android:scheme="https" />
+        <data
+          android:host="novoda.com"
+          android:pathPrefix="/bonfire/welcome"
+          android:scheme="http" />
```
joetimmins: Do we definitely need two `<data>` tags? I presume we do but it would be cooler if we could do `<android:scheme="http|https">` or something

Firebase/android/app/src/main/java/com/novoda/bonfire/link/FirebaseDynamicLinkFactory.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private Uri welcomeDeepLinkFromUser(User user) {
+        return Uri.parse(deepLinkBaseUrl)
+                .buildUpon()
+                .appendPath("welcome")
+                .appendQueryParameter(SENDER, user.getName())
+                .build();
+    }
+
+}
```
joetimmins: How come we have one method returning a java.net.URI and the other returning an android.net.Uri?

joetimmins:
Left a few comments but nothing major 👍

Firebase/android/app/src/main/java/com/novoda/bonfire/link/FirebaseDynamicLinkFactory.java
JozefCeluch: I guess it's because the interface is in core and there is no android there?

Firebase/android/app/src/main/AndroidManifest.xml
JozefCeluch: There could be a separate data tag for each attribute so that the host and pathPrefix aren't duplicated

PR #69, opened 2016-06-02
Firebase/ios refactor

yvettecook:
RxTesting! 🎉 

Added tests for PersistedChannelsService and PersistedChatService/


Firebase/ios/Bonfire/Core/Channel/Channels.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+
+struct Channels {
```
dominicfreeston: As well as being "good practice", turns our first class collections have the advantage of making Rx testing much easier. `Event` is only `Equatable` if the `Element` is also `Equatable`, and arrays can't be `Equatable` (being able to equate two arrays of `Equatable` objects doesn't mean it itself conforms to `Equatable`. First class collections, however, can be `Equatable`, which makes it much easier to verify the results of the test against the list of expected `Recorded<Event<Element>>`.

Also: `Equatable`.

Firebase/ios/Bonfire/Core/Channel/PersistedChannelsService.swift
```diff
@@ -0,0 +1,104 @@
+import Foundation
+import RxSwift
+
+class PersistedChannelsService: ChannelsService {
```
dominicfreeston: These `PersistedServices` match the Java implementations - they push the business of actually writing to the database out into a `Database` protocol, which allows us to test that we're writing to all the relevant parts of the database (a downside of a synced database like Firebase is the requirement for intimate knowledge of how things are stored on the part of the client, but we can at least make sure it's well defined and tested).

Firebase/ios/Bonfire/Core/Chat/ChatPresenter.swift
```diff
@@ -35,8 +35,12 @@ final class ChatPresenter {
         }).addDisposableTo(disposeBag)
 
         chatService.chat(channel).subscribe(
-            onNext: { [weak self] chat in
-                self?.chatDisplayer.display(chat)
+            onNext: { [weak self] result in
+                switch result {
+                case .Success(let chat):
+                    self?.chatDisplayer.display(chat)
+                default: break
```
dominicfreeston: We should probably do some kind of error handling here, show a message or something.

Firebase/ios/Bonfire/Mobile/Utilities/FirebaseConvertible.swift
```diff
@@ -1,8 +1,20 @@
...
+import Firebase
 
 protocol FirebaseConvertible {
+    init(snapshot: FIRDataSnapshot) throws
     init(firebaseValue: AnyObject) throws
     func asFirebaseValue() -> AnyObject
 }
 
+extension FirebaseConvertible {
+    init(snapshot: FIRDataSnapshot) throws {
```
dominicfreeston: This init makes it easier to add the conversion to a typed model object into the rx pipeline. The only caveats is sometimes `FIRDataSnapshot` come out typed as `AnyObject`, which means the `firebaseValue` initialiser gets called instead, so some care is required.

Firebase/ios/BonfireTests/Chat/PersistedChatServiceTests.swift
```diff
@@ -0,0 +1,105 @@
...
+        // When
+        chatService.chat(channel).subscribe(testableObserver)
+        scheduler.start()
+
+        // Then
+        let expectedEvents = [
+            next(0, DatabaseResult.Success(chat))
+        ]
+
+        XCTAssertEqual(testableObserver.events, expectedEvents)
```
dominicfreeston: I like this quite a bit - being able to define clearly the events you expect is neat. RxTests FTW.

Firebase/ios/BonfireTests/PersistedChannelServiceTests.swift
```diff
@@ -0,0 +1,253 @@
...
+    var testableUsersObserver: TestableObserver<Users>!
+    var mockChannelDatabase: ChannelsDatabase!
+    var mockUserDatabase: UserDatabase!
+    var channelService: PersistedChannelsService!
+
+    let testUser1 = User(name: "TestUser1", id: "1", photoURL: nil)
+    let testUser2 = User(name: "TestUser2", id: "2", photoURL: nil)
+
+    let publicChannel = Channel(name: "💣", access: .Public)
+    let privateChannel = Channel(name: "🙈", access: .Private)
```
dominicfreeston: From now on, all tests shall have emojis, especially monkey emojis.

Firebase/ios/BonfireTests/PersistedChannelServiceTests.swift
```diff
@@ -0,0 +1,253 @@
+import XCTest
+import RxSwift
+import RxTests
+@testable import Bonfire
+
+
+class Remembrancer {
```
dominicfreeston: We should probably move this stuff out into its own file. Also some context on what the hell this does.

Basically it's away to pretend-mock and keep track of method calls so we can check the call stack at the end of the test. It's far from perfect, but I think there's something in the idea. Or let's figure out how to write code where Mocks aren't needed 😆

Firebase/ios/BonfireTests/PersistedChannelServiceTests.swift
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+
+    func testThatItCallsTheRightThingsForRemovingOwnerFromChannel() {
+        channelService.removeOwner(testUser1, fromPrivateChannel: privateChannel).subscribe().addDisposableTo(disposeBag)
+
+        let expectedChannel = privateChannel
+        let expectedMethodStack = [
+            MethodCall(identifier: "ChannelsDatabase - removeOwnerFromPrivateChannel", arguments: [testUser1, expectedChannel]),
+            MethodCall(identifier: "ChannelsDatabase - removeChannelFromUserPrivateChannelIndex", arguments: [testUser1, expectedChannel])
```
dominicfreeston: I'd prefer this stuff wasn't stringly-typed, but not sure how best to do it in Swift. It's at times like these I miss the dynamism of Objective-C...

Firebase/ios/BonfireTests/TestScheduler+Extensions.swift
```diff
@@ -0,0 +1,18 @@
...
+
+extension TestScheduler {
+    func singleEventAndComplete<Element>(element: Element) -> Observable<Element> {
+        return self.createColdObservable([
+            next(0, element),
+            completed(0)
+            ]).asObservable()
+    }
+
+    func singleEventAndHang<Element>(element: Element) -> Observable<Element> {
```
dominicfreeston: Do we need to rename this? Or is it descriptive enough? Basically want to simulate a hot observable that stays open because more data might come out later (like when we observe a node in Firebase).

Firebase/ios/BonfireTests/PersistedChannelServiceTests.swift
amlcurran: did whoever create this class read Garth Nix by any chance?

PR #70, opened 2016-06-03
Crash reporting

JozefCeluch:
This adds the Firebase crash reporting and also non-fatal error logging just to try what it looks like.

Enabling the crash reports is as simple as adding the `firebase-crash` dependency and it somehow works automatically so all the rest is logging of non-fatal errors.


Firebase/android/app/build.gradle
```diff
@@ -44,11 +44,12 @@ dependencies {
...
-    compile 'com.google.firebase:firebase-database:9.0.0'
-    compile 'com.google.firebase:firebase-auth:9.0.0'
-    compile 'com.google.firebase:firebase-invites:9.0.0'
-    compile 'com.google.android.gms:play-services-auth:9.0.0'
+    compile 'com.google.firebase:firebase-core:9.0.1'
+    compile 'com.google.firebase:firebase-database:9.0.1'
+    compile 'com.google.firebase:firebase-auth:9.0.1'
+    compile 'com.google.firebase:firebase-crash:9.0.1'
+    compile 'com.google.firebase:firebase-invites:9.0.1'
+    compile 'com.google.android.gms:play-services-auth:9.0.1'
```
joetimmins: 👍

joetimmins:
This all looks pretty straightforward - add the ErrorLogger in the dependencies, pass it into the presenters, and use it to log an error if there is one. 👍

Firebase/android/core/src/main/java/com/novoda/bonfire/analytics/ErrorLogger.java
```diff
@@ -0,0 +1,7 @@
+package com.novoda.bonfire.analytics;
+
+public interface ErrorLogger {
+
+    void reportError(String message, Throwable throwable);
```
tasomaniac: The trend on this is to put `Throwable` in the 1st param. 
It is because if you want to convert the message into a format with additional varags, the last parameter will be `Object...`

JozefCeluch: good point

frapontillo:
LGTM 👍 leaving the merge to the firebase team.

tasomaniac:
Wow, firebase team. So cool. :)

PR #71, opened 2016-06-03
Firebase/feature/test database classes

joetimmins:
This PR adds unit tests for the Firebase database classes that are in the `app` module. We wanted to get some test coverage around this for three reasons - to catch any bugs in the code we use to talk to Firebase, to enable us to refactor more safely and to improve the design of our classes by making them more testable.

In order to do this some static methods that set up the various different Firebase database listeners were converted to instance methods on the FirebaseObservableListeners class. This was purely because Mockito doesn't allow mocking of static methods.

Once it was possible to inject some mock values for the FirebaseObservableListeners, the return values could be stubbed out, either using implementations of Mockito's `Answer` interface or simple return values. Across the three classes that depend on `FirebaseDatabase` there were some common patterns in setting up the necessary test dependences so these have been moved into the `FirebaseTestHelpers` class.


JozefCeluch:
👍  LGTM

frapontillo:
Just a question: why did you use `TestObserver` instead of `TestSubscriber`?

But LGTM 👍

joetimmins:
We never need to unsubscribe in the tests, so I didn't see much point using the `TestSubscriber`. Observer is just a subscriber that can't unsubscribe right? Still an Rx newbie.

Hmmmm maybe we *should* test what happens when we unsubscribe....

frapontillo:
My question was more due to the fact that `TestSubscriber` has many more helper methods to assert stuff, such as `assertValue`, which you could have used instead of the more cumbersome `observer.assertReceivedOnNext(Collections.singletonList(expectedValue));`

joetimmins:
That's a good shout, didn't realise those existed. Don't merge yet, I will use those instead.

joetimmins:
Updated to use TestSubscribers in the FirebaseTestHelpers class. Only had to change that one class instead of all the database test classes :)

frapontillo:
Ship it!

PR #73, opened 2016-06-06
Channel name can only be one emoji

JozefCeluch:
In order to limit the number of channels that users can create we decided to limit channel names to a single emoji. 

Database rules were updated and also error messages in the when the new channel screen is being created. Not all emoji characters are supported on the database (if you're wondering why, just look at the regex that validates the input).

| Before | After |
| --- | --- | 
| channel_no_emoji  | channel_emoji|
| List | Grid |
| channel_list | channels_grid |

@Dorvaryn did the database rules


Firebase/android/app/src/main/java/com/novoda/bonfire/channel/view/ChannelsAdapter.java
```diff
@@ -21,7 +24,8 @@ public void update(Channels channels){
 
     @Override
     public ChannelViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
-        return new ChannelViewHolder(new ChannelView(parent.getContext()));
+        LayoutInflater inflater = LayoutInflater.from(parent.getContext());
```
Dorvaryn: Maybe pass the layout inflater as a param to the adapter.

Firebase/android/app/src/main/java/com/novoda/bonfire/channel/view/NewChannelView.java
```diff
@@ -53,26 +53,55 @@ public void detach(ChannelCreationListener channelCreationListener) {
...
+
+        private boolean inputWasEmpty(int start, int before) {
+            return start == 0 && before == 0;
+        }
+
+        private boolean characterRemoved(int count) {
+            return count == 0;
+        }
+
+        private boolean sequenceIsValidEmoji(CharSequence sequence, int start, int count) {
```
Dorvaryn: Maybe it's time to extract some of that logic.

Dorvaryn: We can probably test this.

Firebase/android/app/src/main/java/com/novoda/bonfire/chat/view/ChatAdapter.java
```diff
@@ -23,7 +25,8 @@ public void update(Chat chat) {
 
     @Override
     public MessageViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
-        return new MessageViewHolder(new MessageView(parent.getContext()));
+        LayoutInflater inflater = LayoutInflater.from(parent.getContext());
```
Dorvaryn: We can pass the inflater as a param in the adapter here as well

Firebase/android/app/src/main/java/com/novoda/bonfire/user/view/UsersAdapter.java
```diff
@@ -25,7 +27,8 @@ public void update(List<UsersView.SelectableUser> users) {
 
     @Override
     public UserViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
-        return new UserViewHolder(new UserView(parent.getContext()));
+        LayoutInflater inflater = LayoutInflater.from(parent.getContext());
```
Dorvaryn: Same here

PR #77, opened 2016-06-08
Snapshot for generating screenshots for the app store

yvettecook:
Using [Snapshot](https://github.com/fastlane/fastlane/tree/master/snapshot) to generate the screenshots for the App Store submission process.

Uses UITests, which is clever, but there are some drawbacks - mainly that you have to ensure you are logged into Bonfire (via Google Auth) on all simulators before running the process.

Also, imagine you could do some clever things with scheme to ensure the UITest that generates the screenshots is only run on the CI.

But, definitely once it's set up, it saves a lot of tedious running the app on ~6 different simulators and screen-grabbing each screen.


amlcurran:
💯 for the screenshots being in VCS (and therefore versioned too)

PR #81, opened 2016-06-09
Firebase/feature/channel list ui

Dorvaryn:
This PR implements the design for the Channel list UI.

| Before | After |
| --- | --- | 
beforedesign|afterdesign


Dorvaryn:
PRB is broken that is the IOS one that launched. This is OK to merge

PR #84, opened 2016-06-09
Firebase/ios/fastlane setup

dominicfreeston:
* Some code cleanup based on SwiftLint feedback.
* Setup fastlane for running tests and beta deployment


dominicfreeston:
Note that the ❌s are due to a failed attempt at setting put jenkins to run the tests using fast lane. I've given up for now as it's not a top priority.

PR #85, opened 2016-06-09
Firebase/feature/chat view ui

Dorvaryn:
This PR implements design for the Chat Screen.

| Before | After |
| --- | --- | 
beforechat|afterchat
beforechattype|afterchattype


Firebase/android/app/src/main/java/com/novoda/bonfire/chat/view/ChatView.java
```diff
@@ -133,4 +139,15 @@ public boolean onMenuItemClick(MenuItem item) {
         }
     };
 
+    private class ChatItemDecoration extends RecyclerView.ItemDecoration {
+        @Override
+        public void getItemOffsets(Rect outRect, View view, RecyclerView parent, RecyclerView.State state) {
+            int horizontalMargin = getResources().getDimensionPixelOffset(R.dimen.list_item_horizontal_margin);
+            int verticalMargin = getResources().getDimensionPixelOffset(R.dimen.list_item_vertical_margin);
```
alexstyl: It would be cleaner to export these two as fields to the decoration class

Firebase/android/app/src/main/java/com/novoda/bonfire/chat/view/MessageView.java
```diff
@@ -28,6 +28,7 @@
 
     public MessageView(Context context, AttributeSet attrs) {
         super(context, attrs);
+        setOrientation(VERTICAL);
```
alexstyl: Since you are forcing the orientation, it might be a good idea to use the `super.setOrientation()` instead, override the `setOrientation()` method and only print a log or throw an exception if someone tries to mess with it

Firebase/android/app/src/main/res/drawable/bubble_shape_grey.xml
```diff
@@ -0,0 +1,37 @@
...
+    android:top="12dp"
+    android:width="12dp"
+    android:height="12dp">
+    <shape android:shape="oval">
+
+      <solid android:color="@color/bubble_grey" />
+
+    </shape>
+  </item>
+</layer-list>
```
alexstyl: This drawable is the 💣 ! 

💯

alexstyl:
Just a couple of minor comments. It has my 👍  when they are resolved :D

One small tip. Since you have a separate bubble for the editor, you don't really need the underline in the EditText. It will look cleaner too.

Firebase/android/app/src/main/java/com/novoda/bonfire/chat/view/ChatView.java
Dorvaryn: 👍

Firebase/android/app/src/main/java/com/novoda/bonfire/chat/view/MessageView.java
Dorvaryn: 👍

PR #88, opened 2016-06-13
Makefest - MVP for simon says spike

blundell:
simon-demo

An MVP that shows a sequence that you need to copy.

Code is gross .. but works.

Needs some aesthetic improvements :-) feel free to get invovled!


xrigau:
I thought the Simon Says game always adds one to the end instead of random every time?

so it'd be:
`1`
`1 1`
`1 1 4`
`1 1 4 2` etc.

xrigau:
https://www.youtube.com/watch?v=4YhVyt4q5HI

blundell:
That is easily changeable, I thought that as well but random works work ( fork it and play :-) ) after you've played let me know if you think it should change!

xrigau:
simon says :shipit:

PR #91, opened 2016-06-13
Add Remote config to trigger sorting of channels

Dorvaryn:
This PR allows for remote configuration of the ordering of the Channels list.

### DIFF
--- a/Firebase/android/app/src/main/java/com/novoda/bonfire/FirebaseConfig.java
+++ b/Firebase/android/app/src/main/java/com/novoda/bonfire/FirebaseConfig.java
@@ -1,0 +1,58 @@
+package com.novoda.bonfire;
+
+import android.support.annotation.NonNull;
+
+import com.google.android.gms.tasks.OnFailureListener;
+import com.google.android.gms.tasks.OnSuccessListener;
+import com.google.firebase.remoteconfig.FirebaseRemoteConfig;
+import com.google.firebase.remoteconfig.FirebaseRemoteConfigSettings;
+import com.novoda.bonfire.analytics.ErrorLogger;
+
+public class FirebaseConfig implements Config {
+
+    private static final String ORDER_CHANNELS_BY_NAME = "orderChannelsByName";
+    private static final int CACHE_EXPIRATION_IN_SECONDS = 3600;
+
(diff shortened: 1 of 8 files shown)


JozefCeluch:
👍 LGTM

Firebase/android/core/src/main/java/com/novoda/bonfire/channel/presenter/ChannelsPresenter.java
```diff
@@ -48,6 +59,7 @@ public void call(Authentication authentication) {
                     }
                 })
                 .flatMap(channelsForUser())
+                .map(sortIfConfigured())
```
friedger: Where will you fetch new values from the server?

Dorvaryn: The new values of the config or of the data ?

friedger: Of the config

Dorvaryn: The Firebase config is handling it automatically, there is a cache time parameter on the init and it will update if the cache is expired.

Firebase/android/app/src/main/java/com/novoda/bonfire/FirebaseConfig.java
```diff
@@ -0,0 +1,58 @@
...
+        FirebaseRemoteConfigSettings configSettings = new FirebaseRemoteConfigSettings.Builder()
+                .setDeveloperModeEnabled(BuildConfig.DEBUG)
+                .build();
+        firebaseRemoteConfig.setConfigSettings(configSettings);
+        firebaseRemoteConfig.setDefaults(R.xml.remote_config_defaults);
+        return new FirebaseConfig(firebaseRemoteConfig);
+    }
+
+    public FirebaseConfig init(final ErrorLogger errorLogger) {
+        firebaseRemoteConfig.fetch(CACHE_EXPIRATION_IN_SECONDS)
```
Dorvaryn: Cache expiry configuration, it will check for new values if the cached value is older than this number.

PR #93, opened 2016-06-13
Update designs for selected screen

Dorvaryn:
This PR updates the design of the add user screen.

| Before | After
| --- | --- |
device-2016-06-13-151914|device-2016-06-13-151756

### DIFF
--- a/Firebase/android/app/src/main/java/com/novoda/bonfire/user/view/UsersView.java
+++ b/Firebase/android/app/src/main/java/com/novoda/bonfire/user/view/UsersView.java
@@ -81,6 +81,7 @@ public class UsersView extends LinearLayout implements UsersDisplayer {
         }
         selectableUsers = usersWithUpdatedSelection;
         usersAdapter.update(selectableUsers);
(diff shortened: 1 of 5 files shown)


Firebase/android/app/src/main/java/com/novoda/bonfire/user/view/UsersView.java
```diff
@@ -81,6 +81,7 @@ public void displaySelectedUsers(Users selectedUsers) {
         }
         selectableUsers = usersWithUpdatedSelection;
         usersAdapter.update(selectableUsers);
+        toolbar.setTitle(selectedUsers.size() + " Selected");
```
JozefCeluch: this could be extracted to a string resource with a parameter

JozefCeluch:
👍  LGTM

Firebase/android/app/src/main/java/com/novoda/bonfire/user/view/UsersView.java
Dorvaryn: There is a few other places where we can do that I'd do a pass later for those.

PR #95, opened 2016-06-13
Update design for login screen

JozefCeluch:
This PR updates the design of the login screen and adds a disclaimer about privacy.

| Before | After
| --- | --- |
bonfire-login | login_after

### DIFF
--- a/Firebase/android/app/src/main/java/com/novoda/bonfire/login/view/LoginView.java
+++ b/Firebase/android/app/src/main/java/com/novoda/bonfire/login/view/LoginView.java
@@ -2,18 +2,18 @@ package com.novoda.bonfire.login.view;
 
 import android.content.Context;
 import android.util.AttributeSet;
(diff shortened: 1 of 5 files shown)


Firebase/android/app/src/main/java/com/novoda/bonfire/login/view/LoginView.java
```diff
@@ -25,6 +25,7 @@ protected void onFinishInflate() {
         super.onFinishInflate();
         View.inflate(getContext(), R.layout.merge_login_view, this);
         loginButton = Views.findById(this, R.id.sign_in_button);
+        loginButton.setSize(SignInButton.SIZE_WIDE);
```
Dorvaryn: Can't be set in xml ?

JozefCeluch: yep 👍

PR #103, opened 2016-06-17
Makefest - new game play

blundell:
Change the game so that it is incremental in changing the pattern.

new-sequence

Next I'll make it go faster as the game goes on


xrigau:
lcn0yf1rclang

PR #105, opened 2016-06-17
Improve welcome screen UI

JozefCeluch:
This PR is just a small update to the existing welcome screen to match the final design.

| Before | After |
| --- | --- |
| welcome_before | welcome_final  |


frapontillo:
:shipit: because 🍺

PR #107, opened 2016-06-20
Draw bubble backgrounds manually

JozefCeluch:
#### Problem

Drawable XMLs that we use for backgrounds are not drawn correctly on SDK version smaller than 23. 

The weight/height attributes of items in the layer-list are ignored and the items are stretched to match the parent size. This made it impossible to create small circles in the message layout and the welcome/login screen layouts.

#### Solution

Remove the XML drawables and draw the backgrounds directly in the onDraw method of the custom views.

#### Screenshots 

| Before | After |
| --- | --- |
| login_before | login_after |
| message_before | message_after |


Firebase/android/app/src/main/res/layout/merge_login_view.xml
```diff
@@ -8,18 +8,19 @@
     android:layout_height="match_parent"
     android:orientation="vertical">
 
-    <FrameLayout
+    <com.novoda.bonfire.view.BubblyLinearLayout
```
rock3r: Why the switch from `FrameLayout` to `LinearLayout` with no changes to the contents? ⚠️

Firebase/android/app/src/main/java/com/novoda/bonfire/view/BubblyLinearLayout.java
```diff
@@ -0,0 +1,78 @@
...
+import android.content.res.TypedArray;
+import android.graphics.Canvas;
+import android.graphics.Paint;
+import android.graphics.RectF;
+import android.util.AttributeSet;
+import android.widget.LinearLayout;
+
+import com.novoda.bonfire.R;
+
+public class BubblyLinearLayout extends LinearLayout {
```
rock3r: Shouldn't be a `LinearLayout` but rather a `FrameLayout` -- see https://github.com/novoda/spikes/pull/107/files#r67866404

rock3r: Ok there's cases where this needs to be a `LinearLayout`, nvm

rock3r:
I think it'd be way easier to create a `BubbleDrawable` instead, and use that for BGs...

Firebase/android/app/src/main/res/layout/merge_login_view.xml
JozefCeluch: The other place where it is used has multiple views inside the layout so extended the layout that would work in both places

JozefCeluch:
@rock3r It would be easier, and we had it like that but it didn't work on APIs smaller than 23 because the width and height of items in layer-list are ignored. The easiest solution for message bubble would be a 9 patch but that wouldn't work for the bubble in the login view anyway.

rock3r:
Maybe I wasn't explaining as best as I could:

```java
public class BubblyDrawable extends Drawable {

  // ...

  @Override
  public void draw() {
    // Your drawing logic here
  }
}
```

Then you just `setBackground(new BubblyDrawable())` :)

JozefCeluch:
Oh yeah, good point. But I would have to set it in the code instead of having everything in the xml, right?

rock3r:
Yup, but it's EZ.

Firebase/android/app/src/main/java/com/novoda/bonfire/login/view/LoginView.java
```diff
@@ -24,6 +25,7 @@ protected void onFinishInflate() {
         super.onFinishInflate();
         View.inflate(getContext(), R.layout.merge_login_view, this);
         loginButton = Views.findById(this, R.id.sign_in_button);
+        Views.findById(this, R.id.login_icon_frame_layout).setBackground(new BubblyDrawable(getContext()));
```
rock3r: What about doing one thing at a time?

```java
View frame = Views.findById(this, R.id.login_icon_frame_layout);
frame.setBackground(new BubblyDrawable(getContext()));
```

Firebase/android/app/src/main/java/com/novoda/bonfire/view/BubblyDrawable.java
```diff
@@ -75,4 +58,18 @@ private void drawBubble(Canvas canvas, int xPosition, int yPosition, int diamete
...
+    }
+
+    @Override
+    public void setColorFilter(ColorFilter colorFilter) {
+        paint.setColorFilter(colorFilter);
+    }
+
+    @Override
+    public int getOpacity() {
+        return PixelFormat.OPAQUE;
```
rock3r: Should be `TRANSLUCENT` unless _every_ pixel within the drawable's bounds is 100% opaque! ⚠️

Firebase/android/app/src/main/java/com/novoda/bonfire/welcome/view/WelcomeView.java
```diff
@@ -34,6 +35,7 @@ protected void onFinishInflate() {
         welcomeMessage = Views.findById(this, R.id.welcome_message_view);
         userAvatar = Views.findById(this, R.id.user_avatar);
         proceedButton = Views.findById(this, R.id.proceed_button);
+        Views.findById(this, R.id.welcome_sender_layout).setBackground(new BubblyDrawable(getContext()));
```
rock3r: Same as before

rock3r:
Couple of comments before LGTM

rock3r:
LGTM now 👍

Dorvaryn:
Should the TextView still be a custom view ? Seems like we could use a custom drawable as well.

rock3r:
That makes sense too @Dorvaryn -- my bad for missing that out. Assumed it'd be in a separate PR, not sure why. Was tired I suppose :D

Firebase/android/app/src/main/java/com/novoda/bonfire/chat/view/ChatAdapter.java
```diff
@@ -30,11 +33,17 @@ public void update(Chat chat, User user) {
 
     @Override
     public MessageViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
+        MessageBubbleDrawable bubbleDrawable;
+        MessageView messageView;
         if (viewType == 0) {
```
rock3r: Could I recommend extracting a `VIEW_TYPE_MESSAGE_OTHER_USERS` maybe? :) Also a `VIEW_TYPE_MESSAGE_THIS_USER` and use it here (`else if` it), so you have an `else` that can throw

Firebase/android/app/src/main/java/com/novoda/bonfire/view/MessageBubbleDrawable.java
```diff
@@ -26,32 +27,15 @@
...
+    public MessageBubbleDrawable(Context context, @ColorRes int color, Gravity gravity) {
+        this.gravity = gravity;
         rect = new RectF();
         paint = new Paint();
         paint.setStyle(Paint.Style.FILL);
         paint.setAntiAlias(true);
-        resolveAttributes(context, attrs);
-        resolveDimensions(getResources());
-    }
+        paint.setColor(context.getResources().getColor(color));
```
rock3r: This, or use `setColorFilter()` to do the tinting:

```java
ColorFilter colorFilter = new PorterDuffColorFilter(myColor, PorterDuff.Mode.SRC_IN);
setColorFilter(colorFilter);
```

tasomaniac:
Looks good but just 1 comment. Wouldn't this be very easy with 9-patch as well?

Firebase/android/app/src/main/java/com/novoda/bonfire/view/MessageBubbleDrawable.java
JozefCeluch: Does setting a color have any disadvantages compared to setting a color filter?

JozefCeluch:
@tasomaniac Yes, for the messages 9-patch would work

rock3r:
@tasomaniac https://github.com/novoda/spikes/pull/107#issuecomment-227441858

Firebase/android/app/src/main/java/com/novoda/bonfire/view/MessageBubbleDrawable.java
rock3r: Nah, it's just an FYI, I don't mind either way :)

tasomaniac:
Oh, I read the comments but totally missed that part of @JozefCeluch 's comment. 👍

PR #108, opened 2016-06-22
Firebase/death to whitespace

dominicfreeston:
New features:

* We now disable the send button for empty messages, including messages made of only whitespace.
* We trim the string before submitting it.

Refactoring / clean up:
* Removed unused property and import of RxSwift in AppDelegate
* Now using traditional target/action rather than `rx_` extensions in all view controllers, for consistency
* Removed some unused lines and added a little description

### DIFF
--- a/Firebase/ios/Bonfire/Mobile/AppDelegate.swift
+++ b/Firebase/ios/Bonfire/Mobile/AppDelegate.swift
@@ -12,8 +12,10 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
         FIRApp.configure()
         GIDSignIn.sharedInstance().clientID = FIRApp.defaultApp()?.options.clientID
 
-//                try! FIRAuth.auth()?.signOut()
-//                GIDSignIn.sharedInstance().signOut()
+        // Uncomment these lines to sign out and start afresh!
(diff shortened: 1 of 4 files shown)


Firebase/ios/Bonfire/Mobile/AppDelegate.swift
```diff
@@ -1,21 +1,21 @@
...
-    let disposeBag = DisposeBag()
     var userService: UsersService!
 
     func application(application: UIApplication, didFinishLaunchingWithOptions launchOptions: [NSObject: AnyObject]?) -> Bool {
         FIRApp.configure()
         GIDSignIn.sharedInstance().clientID = FIRApp.defaultApp()?.options.clientID
 
-//                try! FIRAuth.auth()?.signOut()
-//                GIDSignIn.sharedInstance().signOut()
+        // Uncomment these lines to sign out and start afresh!
```
Dorvaryn: Do we want to keep this in ?

dominicfreeston: I was wondering - we have no logout functionality, so if you want to play with the app you might want to know how to do it?

I'm not precious about it though.

PR #110, opened 2016-06-27
Avoid cancelling fetch when unsubscribing from the data

Dorvaryn:
When unsubscribing from the tasks the behaviour was bound to the scheduler and any init still running would be cancelled.

We now allow the init fetch to continue running and cache it's result so that when the next client subscribes the data is there.

### DIFF
--- a/Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
+++ b/Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
@@ -8,10 +8,11 @@ import com.novoda.todoapp.tasks.data.model.Tasks;
 
 import java.util.concurrent.TimeUnit;
 
+import rx.Notification;
 import rx.Observable;
 import rx.Subscriber;
+import rx.functions.Action1;
 import rx.functions.Func0;
-import rx.schedulers.Schedulers;
 
 public final class InMemoryLocalTaskDataSource implements LocalTasksDataSource {
 
@@ -54,7 +55,7 @@ public final class InMemoryLocalTaskDataSource implements LocalTasksDataSource {
                     return Observable.just(localTasks);
                 }
             }
-        }).delay(localDelay, localDelayUnit, Schedulers.immediate());
+        }).compose(this.<Tasks>delay());
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
```diff
@@ -78,7 +79,7 @@ public void call(Subscriber<? super SyncedData<Task>> subscriber) {
                 subscriber.onNext(taskSyncedData);
                 subscriber.onCompleted();
             }
-        }).delay(localDelay, localDelayUnit, Schedulers.immediate());
```
Dorvaryn: The existing delay logic does not retry on interruption.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryRemoteTaskDataSource.java
```diff
@@ -150,7 +150,25 @@ public void call(Task task) {
                 dataSourceTasks.remove(taskId);
                 return Observable.empty();
             }
-        }).delay(remoteDelay, remoteDelayUnit, Schedulers.immediate());
+        }).compose(this.<Void>delay());
+    }
+
+    private <T> Observable.Transformer<T, T> delay() {
```
eduardb: This kind of looks like a duplicate of `InMemoryLocalTaskDataSource.delay()`, maybe it can be extracted to a class?

Dorvaryn: I tend to follow the rule of three, duplicate once refactor when you have 3 instances.
In this case it is a class mocking behaviour, I would probably leave it at that. If we have the need for a 3rd then we can refactor.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
```diff
@@ -89,6 +90,24 @@ public void call(Subscriber<? super Void> subscriber) {
                 localTasks = Tasks.empty();
                 subscriber.onCompleted();
             }
-        }).delay(localDelay, localDelayUnit, Schedulers.immediate());
+        }).compose(this.<Void>delay());
+    }
+
+    private <T> Observable.Transformer<T, T> delay() {
```
joetimmins: 👍

PR #117, opened 2016-07-01
#eNews node v2

ouchadam:
Problem -

Parsing enews was failing due to a slack room being created the same name, enews was being replaced with #C0YNBKANM

There was also some extra information that is useful to attach the the messages, such as the posters name and the news link, title and image.

Solution 

- Filter messages by including the slack room id
- linking message userIds to to Users to give us the message posters real name

```
{ 
  originalMessage: '<#C0YNBKANM|enews> A giant floating barrier might be the answer to cleaner     
  oceans \n<http://yahoonewsdigest-us.tumblr.com/146517763724>',
  title: 'A giant floating barrier might be the answer to cleaner oceans',
  link: 'http://yahoonewsdigest-us.tumblr.com/146517763724',
  poster: 'Wagner Truppel',
  imageUrl: 'https://msearch.zenfs.com/atomed/b26f4d10-3bb7-11e6-af21-6c3be5be5218.jpg'
}
```

### DIFF
--- a/enews-node/app.js
+++ b/enews-node/app.js
@@ -6,20 +6,12 @@ var newsFetcher = new NewsFetcher(token);
 var latestDate = new Date();
 var oldestDate = new Date();
 
-oldestDate.setDate(latestDate.getDate() - 31);
+oldestDate.setDate(latestDate.getDate() - 7);
 
 // node timestamps are in milliseconds, need to convert to epoch
 var latest = latestDate / 1000;
 var oldest = oldestDate / 1000;
 
-// newsFetcher.getEnews(oldest, latest, function(eNews) {
-//   eNews.forEach(function(each) {
-//     console.log(each);
-//   });
-// });
-
-newsFetcher.getEnews(oldest, latest, function(messages) {
-  messages.forEach(function(each) {
-    console.log(each);
-  });
+newsFetcher.getEnews(oldest, latest, function(eNews) {
+  console.log(eNews);
 });
--- a/enews-node/app.js
+++ b/enews-node/app.js
@@ -12,8 +12,14 @@ oldestDate.setDate(latestDate.getDate() - 31);
 var latest = latestDate / 1000;
 var oldest = oldestDate / 1000;
 
-newsFetcher.getEnews(oldest, latest, function(eNews) {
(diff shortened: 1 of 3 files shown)


enews-node/newsFetcher.js
```diff
@@ -1,72 +1,70 @@
...
+    userIds.forEach(function(userId) {
+      promises.push(getUser(userId));
+    });
+    return promises;
   }
 
-  function contains(input, check) {
-    return input.indexOf(check) > -1;
+  function getUser(userId) {
+    return Q.promise(function(resolve, reject, notify) {
```
ouchadam: the magic is real 🎉

enews-node/app.js
```diff
@@ -6,14 +6,12 @@ var newsFetcher = new NewsFetcher(token);
 var latestDate = new Date();
 var oldestDate = new Date();
 
-oldestDate.setDate(latestDate.getDate() - 31);
```
ouchadam: get the last 7 days in the usage example instead of 31

blundell:
Do you think it's sensible having problems names? Anon

ouchadam:
I don't think it's a bad thing to have here, we don't need to propagate it into the html, but I see no reason to not fetch it at this level as well

enews-node/newsFetcher.js
```diff
@@ -1,72 +1,70 @@
...
+      var eNews = toEnews(eNewsMessages, users);
+      callback(eNews);
     });
-    return enews;
   }
 
   function isEnewsMessage(message) {
-    return isValidMessage(message) && isEnews(message.text)
+    var messageText = message.text;
+    return contains(messageText, '#enews') || contains(messageText, '#eNews') || contains(messageText, '#C0YNBKANM');
```
mr-archano: so is this `#C0YNBKANM` temporary?

blundell:
Ya ya ya :+1: any more comments can be addressed inline right. Ship it to get the eeemail

enews-node/newsFetcher.js
ouchadam: it's the redirect slack creates for `#enews` channel, for now I guess it's a constant

PR #118, opened 2016-07-02
BuildPropertiesPlugin: Add bintray support

mr-archano:
What the tin says. Follow-up would be releasing the plugin via the Novoda CI.


xrigau:
:+1:

PR #120, opened 2016-07-03
BuildProertiesPlugin: Fix maven coordinates

mr-archano:
Version `0.9.5` of the `build-properties-plugin` has been released to Bintray with the wrong coordinates:
```
  <groupId>com.novoda.build-properties</groupId>
  <artifactId>build-properties</artifactId>
  <version>0.9.5</version>
```

We want to align our artifacts to always use `com.novoda` as groupId, removing the redundant artifactId from it. This leads to the deploy of version `0.9.6`. The new pom will look like:
```
  <groupId>com.novoda</groupId>
  <artifactId>build-properties-plugin</artifactId>
  <version>0.9.6</version>
```


ouchadam:
💯

PR #121, opened 2016-07-03
enews-letter html generation and github page deployment

ouchadam:
Problem -

We want to take the enews parsed from slack and turn it into a newsletter

Solution -

- moved all of the enews code into a an enews directory

- Include the local enews-node module and create a bunch of html!

- small script to take the output html and force push it to the gh-pages branch. This is wrapped inside the npm scripts.

npm install will set up all the dependencies 
npm start will grab the enews from slack and generate the html
npm deploy will push to gh-pages

Usage -

```
cd html-generator-node
npm install && npm start && npm run deploy
```

deployed page : http://novoda.github.io/spikes/


xrigau:
:100:

mr-archano:
Dat font. :trollface:

PR #122, opened 2016-07-03
#enews mailer

ouchadam:
Emails the generated enews html using the provided smtp credentials and comma separated recipients.  

process.env.SMTP_USER;
process.env.SMTP_PASSWORD;
process.env.SMTP_HOST;
process.env.RECIPIENTS;

- Tidied up the enews modules to be proper libs with demo.js


xrigau:
Noice!

enews/mailer-node/enews-mailer.js
```diff
@@ -0,0 +1,48 @@
...
+    secure: true,
+    auth: {
+        user: SMTP_USER,
+        pass: SMTP_PASSWORD
+    }
+  };
+
+  var transporter = nodemailer.createTransport(smtpConfig);
+  var mailOptions = {
+    from: '"Hal Berto 👥" <enews@novoda.com>',
```
blundell: Should this be a property?

ouchadam: could be, it's not a real address, it'll default to the passed in smtp user

PR #126, opened 2016-07-06
[Spike] MonkeyTrap.

xrigau:
Add app to put an overlay on top of the Notification Bar so that it is not accessible to the Monkey UI Exerciser.

This will prevent the Monkey from disconnecting a device from wifi,
enabling airplane mode, etc.

See [README.md](https://github.com/novoda/spikes/blob/spike/monkey-notification-bar-overlay/MonkeyTrap/README.md) for more details

Result:

Without | With
--- | ---
without | with


MonkeyTrap/app/src/main/java/com/novoda/monkeytrap/OverlayView.java
```diff
@@ -0,0 +1,45 @@
...
+import android.view.WindowManager;
+import android.widget.FrameLayout;
+
+class OverlayView extends FrameLayout {
+
+    private static final int SAFETY_MARGIN = 20;
+
+    static WindowManager.LayoutParams createLayoutParams(Resources resources) {
+        final WindowManager.LayoutParams params =
+                new WindowManager.LayoutParams(MATCH_PARENT, retrieveStatusBarHeight(resources) + SAFETY_MARGIN,
```
florianmski: Why do we need a `SAFETY_MARGIN`?

MonkeyTrap/app/src/main/java/com/novoda/monkeytrap/OverlayService.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        instance = this;
+        windowManager = (WindowManager) getSystemService(WINDOW_SERVICE);
+
+        overlayView = new OverlayView(this);
+        overlayView.setBackgroundColor(Color.parseColor("#2FA1D6"));
```
florianmski: I see what you did there 👀 
Having the overlay transparent would look better though, no?

florianmski:
Awesooooooooooome
What are the main differences between this and the project simiasque linked in the README?

MonkeyTrap/app/src/main/java/com/novoda/monkeytrap/OverlayView.java
xrigau: for safety :trollface: 

I imagine it is because the touch are is slightly larger than the visible area of the notifications bar. I took it from https://github.com/Orange-OpenSource/simiasque (as the README says)

xrigau:
@florianmski that project uses the AndroidAnnotations library and I find that a bit gross https://github.com/Orange-OpenSource/simiasque/blob/master/app/build.gradle#L27-L28

The approach is slightly different as well: they use a Broadcast receiver that fires an IntentService that will either show or hide the Overlay and there's a ViewHolder that holds the Overlay view in memory, whereas we just have a Service that holds the Overlay and it changes its visibility in `onStartCommand`

Also, I thought if we wanted to use this internally it'd be safer to be in control of the source code (plus we can add other functionality if we need, like preparing the device before running tests, and the likes)

MonkeyTrap/app/src/main/java/com/novoda/monkeytrap/OverlayService.java
xrigau: Novoda branding 🎉

xrigau: also gives a visible hint which is nice (so that you don't leave it on - to then forget about it - imagine how troll that would be :trollface: )

MonkeyTrap/app/src/main/java/com/novoda/monkeytrap/OverlayView.java
florianmski: Ok but then because the touch area is smaller I guess the monkey will tap less often on the action bar menu items :/
Guess we can't do much about that...

MonkeyTrap/app/src/main/java/com/novoda/monkeytrap/OverlayService.java
florianmski: TRUE DAT 🎉

MonkeyTrap/README.md
```diff
@@ -0,0 +1,36 @@
+# Monkey Trap
+
+A way to prevent the Android UI/Application Exerciser Monkey from accessing the System Notifications tray. This makes sure the Monkey doesn't leave a device without Internet after it finishes.
```
blundell: > device without network connection after it finishes.

MonkeyTrap/app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,31 @@
...
+<manifest package="com.novoda.monkeytrap"
+    xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" />
+
+    <application
+        android:allowBackup="true"
+        android:icon="@mipmap/ic_launcher"
+        android:label="@string/app_name"
+        android:supportsRtl="true">
```
blundell: does it?

MonkeyTrap/app/src/main/java/com/novoda/monkeytrap/OverlayService.java
blundell: write *NOVODA* on the middle of it 💃

MonkeyTrap/app/src/main/AndroidManifest.xml
xrigau: yeah! we don't do much in the UI so it should work

MonkeyTrap/.gitignore
```diff
@@ -0,0 +1,7 @@
+*.iml
```
xrigau: 👍  yeah sorry!

MonkeyTrap/app/src/main/java/com/novoda/monkeytrap/OverlayService.java
xrigau: was thinking of doing it but didn't bother, can do in a later PR (or even add the novoda logo 😄 )

blundell:
LGETM

florianmski:
LGTM2 💃 

:shipit:

xrigau:
note to self: the novoda blue is `#30a4d9`

blundell:
we should write that everywhere

MonkeyTrap/app/src/main/java/com/novoda/monkeytrap/OverlayService.java
xrigau: #129

PR #130, opened 2016-07-07
Implementing a Delete Task button

joetimmins:
#### Problem
Our todo app had no way of deleting a task. The Google examples had a `DELETE TASK` button on their task detail screen.

#### Solution
We added a `DELETE TASK` button in the same place as Google added theirs.
However, ours needed to work slightly differently. The Google version just deletes the task from their server, goes back a screen, and hopes it has been deleted quickly enough not to appear in their main screen. Right now, our app doesn't run its `Service` on a background thread, so it waits on the task detail screen until it has been deleted from the remote backend, then goes back a screen. This is a bit janky, but it'll be fixed in later PRs.

##### Test(s) added 
_yes_ - Tests for the new functionality in `TaskPresenter` and `PersistedTaskService` were added. I also cleaned up some of the existing tests in `PersistedTaskService` to make them a bit easier to read.

##### Screenshots
| Before | Desired | After |
| ------ | ----- | ----- |
| 
no-delete-task| google-delete-task | novoda-delete-task |

###### Paired with 
@Dorvaryn


Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
```diff
@@ -53,9 +53,13 @@ public boolean isEmpty() {
...
     }
 
-    public boolean isMostRecentAction(SyncedData<Task> syncedData) {
-        SyncedData<Task> currentSyncData = internalMap().get(syncedData.data().id());
-        return currentSyncData == null || syncedData.lastSyncAction() >= currentSyncData.lastSyncAction();
+    public boolean hasNoTaskMoreRecentThan(SyncedData<Task> syncedData) {
+        return hasNoTaskMoreRecentThan(syncedData.data().id(), syncedData.lastSyncAction());
+    }
+
+    public boolean hasNoTaskMoreRecentThan(Id id, long lastSyncAction) {
```
Dorvaryn: The name is a bit confusing, it does not check that it does not have any Task more recent than the given timestamp. Rather that there is no action on the given task that is more recent than the given timestamp.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
```diff
@@ -86,6 +90,15 @@ public Tasks onlyCompleted() {
         return Tasks.from(internalMap().filter(isCompleted()));
     }
 
+    public Tasks withoutTask(final Task task) {
```
Dorvaryn: Isn't it basically `delete` or `remove`

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
```diff
@@ -107,4 +120,13 @@ public boolean apply(Map.Entry<Id, SyncedData<Task>> input) {
...
         };
     }
+
+    @Override
+    public String toString() {
+        String result = "";
+        for (Id id : internalMap().keySet()) {
+            result = result + ", " + "{" + id + "; " + internalMap().get(id).syncState() + "}";
+        }
+        return result.substring(2);
```
Dorvaryn: Why the sub string ?

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
```diff
@@ -187,28 +187,28 @@ public Action0 clearCompletedTasks() {
...
             }
         };
     }
 
-    private static Observable.Transformer<List<Task>, Tasks> clearLocallyThenConfirmOrRevert(
-            final Tasks tasksFromEvent,
-            final Tasks uncompletedTasks,
+    private static Observable.Transformer<List<Task>, Tasks> confirmLocalDeletionsOrRevert(
+            final Tasks allLocalTasks,
+            final Tasks uncompletedLocalTasks,
```
Dorvaryn: `uncompletedLocalTasks` is `activeLocalTasks`

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
joetimmins: Previously it was hard to tell if it returned true if the `Tasks` contained the more recent task, or if the `SyncedData` passed in was more recent. `Tasks.isMostRecentAction(syncedData)` ... which is more recent, the `Tasks` or the `syncedData` ? It's confusing.

Perhaps `noActionMoreRecenThan` would be better.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
joetimmins: `remove` would be better, yeah 👍

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
joetimmins: So it chops off the first comma and space

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
joetimmins: 👍

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
Dorvaryn: Sounds good, less confusing for both aspects.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
Dorvaryn: Ah ok

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
joetimmins: 👍 fixed in https://github.com/novoda/spikes/pull/130/commits/27928b970290dc85b8eede7f23cda93d8c04ce83

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
joetimmins: Fixed in https://github.com/novoda/spikes/pull/130/commits/27928b970290dc85b8eede7f23cda93d8c04ce83

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
frapontillo: It'd be indeed clearer if you extract `", "` to a final string and use its length, so that intentions are explicit.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
```diff
@@ -301,6 +306,104 @@ public void call() {
...
+        return asOrchestratedAction(new DataOrchestrator<Void, Event<Tasks>>() {
+            @Override
+            public Event<Tasks> startWith() {
+                Event<Tasks> currentState = taskRelay.getValue();
+                return currentState.asLoadingWithData(markTaskAsDeletedLocallyAt(currentState, task, clock.timeInMillis()));
+            }
+
+            @Override
+            public Event<Tasks> onConfirmed(Void value) {
+                return onError();
```
frapontillo: Why is there an error when it's confirmed?

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
Dorvaryn: 💯

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
Dorvaryn: We don't expect a value back there is two way you can get a confirmation, termination of the stream or termination of the stream with a value. In this case we never expect a value so it would be an error. Some other usages of this will be reversed where they need a value and a confirmation without one would be an error.

frapontillo:
LGTM 👍

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
joetimmins: Fixed in https://github.com/novoda/spikes/pull/130/commits/9e4dbc3c496d02aedcb9de59d79690304f3f4bd5

I kept the delimiter string inside the `toString()` method - felt a bit cleaner not having it at the top of the class as a constant, because oonly the `toString()` method uses it.

Architecture/TodoApp/android/mobile/src/main/res/layout/merge_task_detail_view.xml
```diff
@@ -8,10 +8,11 @@
...
     android:orientation="vertical">
 
-    <com.novoda.todoapp.views.TodoAppBar
-      android:id="@+id/app_bar"
+    <com.novoda.todoapp.views.TodoAppBar xmlns:todoApp="http://schemas.android.com/apk/res/com.novoda.todoapp"
+      android:id="@+id/app_bar_with_delete_button"
       android:layout_width="match_parent"
-      android:layout_height="wrap_content" />
+      android:layout_height="wrap_content"
+      todoApp:mergeLayoutId="@layout/merge_app_bar_with_delete_button" />
```
danybony: Out of curiosity, why did you use the namespace `todoApp`? Wan't OK to use `app`?

joetimmins: Yeah we already use `app` as a namespace on line 3 of this file. It's for the `res-auto` stuff we need for the FAB further down the layout.

danybony:
LGTM 👍

Architecture/TodoApp/android/mobile/src/main/res/layout/merge_task_detail_view.xml
joetimmins: Actually @danybony and I figured out you don't need the custom namespace and `res-auto` will let you use custom attributes just fine.

PR #132, opened 2016-07-09
Build properties plugin: Remove local deploy

mr-archano:
Following the suggestion from @devisnik I removed the embedded maven repo and made the sample app build the sources in `plugin/` as part of its build script.


mr-archano:
@hal9002 test this please.

BuildPropertiesPlugin/sample/app/build.gradle
```diff
@@ -1,15 +1,11 @@
...
-  }
   dependencies {
     classpath 'com.android.tools.build:gradle:2.1.2'
-    classpath "${pluginGroup}:${pluginId}:${pluginVersion}"
   }
 }
 
 apply plugin: 'com.android.application'
-apply plugin: 'build-properties'
+apply plugin: com.novoda.buildproperties.BuildPropertiesPlugin
```
tasomaniac: This is long :) 
I think just `com.novoda.buildproperties` would be cool. But I guess this didn't change in this PR at all, right?

tasomaniac:
I see you had a fight with @stefanhoth in this PR 😸

BuildPropertiesPlugin/sample/app/build.gradle
mr-archano: This is not the standard way of applying the plugin (note the lack of ''). The sample is now compiling against the plugin sources directly, therefore we need to use the class name instead of the string identifier.

tasomaniac: I see. The fully qualified name as a plug-in name is also recommended. Like 'com.android.application'  That's why I thought you changed it. 
:+1:

BuildPropertiesPlugin/plugin/build.gradle
```diff
@@ -1,36 +1,29 @@
...
   }
 }
 
+repositories {
+  jcenter()
+}
+
 apply plugin: 'groovy'
-apply from: rootProject.file('gradle/publish.gradle')
+apply from: file('../gradle/publish.gradle')
```
ouchadam: might be safer to use `rootDir`, not a blocker

ouchadam:
the less buildscript the better 👍

BuildPropertiesPlugin/plugin/build.gradle
mr-archano: You're right, this might be a left-over from some previous (failed) experiment. I'll check if that would work 👍

PR #133, opened 2016-07-10
Build properties plugin: Introducing Entries abstraction

mr-archano:
This PR introduce a level of abstraction handling the access of property values inside `BuildProperties`. The main advantage of such indirection is to define a clear entry point for extending the plugin and provide different ways of retrieving properties (eg: from a zip file, from a maven artifact, etc). 
A `BuildProperties` instance is needed by the plugin extension to index a property set, but doesn't need to hold anymore details about the specific implementation like the `File` we're using or the `Properties` instance we're accessing the values from. It now only holds an instance of `Entries` which is in turn delegated to hold and access all the properties.
At the moment the only implementation provided is `FilePropertiesEntries`, backed by a `Properties` file.


BuildPropertiesPlugin/plugin/build.gradle
```diff
@@ -12,7 +12,7 @@ repositories {
 }
 
 apply plugin: 'groovy'
-apply from: file('../gradle/publish.gradle')
+apply from: rootProject.file('gradle/publish.gradle')
```
ouchadam: 👍

BuildPropertiesPlugin/plugin/src/test/groovy/com/novoda/buildproperties/BuildScriptTest.groovy
```diff
@@ -16,44 +16,41 @@ public class BuildScriptTest {
...
   }
 
   @Test
   public void shouldNotApplyPluginWhenAndroidPluginNotApplied() {
-    Project project = ProjectBuilder.builder()
-            .withProjectDir(temp.newFolder())
-            .build()
     try {
-      plugin.apply(project)
+      project.apply plugin: BuildPropertiesPlugin
```
ouchadam: 👍

BuildPropertiesPlugin/plugin/src/test/groovy/com/novoda/buildproperties/SampleProjectTest.groovy
```diff
@@ -91,7 +91,7 @@ public class SampleProjectTest {
       releaseBuildConfig = new File(PROJECT.projectDir, 'app/build/generated/source/buildConfig/release/com/novoda/buildpropertiesplugin/sample/BuildConfig.java')
       debugResValues = new File(PROJECT.projectDir, 'app/build/generated/res/resValues/debug/values/generated.xml')
       releaseResValues = new File(PROJECT.projectDir, 'app/build/generated/res/resValues/release/values/generated.xml')
-      secrets = BuildProperties.create(new File(PROJECT.projectDir, 'properties/secrets.properties'))
+      secrets = FilePropertiesEntries.create(new File(PROJECT.projectDir, 'properties/secrets.properties'))
```
ouchadam: 👍 makes sense to have this as the api

PR #134, opened 2016-07-10
BuildPropertiesPlugin: Override properties from command line

mr-archano:
Sometimes you want to override a property from command line,  to provide a value that otherwise is specified in a properties file. Assuming you want to use a different value of the property `myProperty` at build time you need only to specify the new value via `-PmyProperty=newValue` and the plugin will ensure that value will be used instead of the one read from the properties file.


ouchadam:
out of interest what happens if you have properties in gradle home, a build properties file, environmental var and command params, who wins?

mr-archano:
In that case the standard gradle resolution strategy applies (not sure about the order of importance). The plugin just expects the new value to end up in the project properties, the how is not relevant.

ouchadam:
👍

PR #135, opened 2016-07-10
BuildPropertiesPlugin: Introduce recursive include support for properties

mr-archano:
We found useful to have properties files that can recursively include other properties files specified in a `include` property. This can help composing set of properties in a hierarchical way allowing to define cohesive sets of properties and helping reusing them. Inherited properties can be overridden by the including set, just redefine the property in the file and its value will be used instead of the one from the included set.


ouchadam:
👍

PR #139, opened 2016-07-11
Avoid Updating the tasks when the value is outdated.

Dorvaryn:
#### Problem
When clearing the completed tasks, if there was a task marked as completed just before we could sometime have the task reappearing as the reply form the sever would override the local state.

#### Solution
We can mark the timestamp of the clearCompleted tasks action and avoid any updates when timestamps are not showing we are the most recent action.
To achieve this a few other refactoring were needed to simplify the code, one notable one is the addition of the scheduler inside the service. This allows for finer grain concurrency decisions in the domain that single threaded actions could not achieve.

##### Test(s) added 
_yes_ - `PersistedTasksService` now has tests covering updates when calling `clearCompletedTasks`

##### Screenshots
| Before | After |
| ------ | ----- |
|before|after|

###### Paired with 
No-one


Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/task/TaskActionListener.java
```diff
@@ -26,7 +26,7 @@ public void save(Optional<String> title, Optional<String> description) {
                     .title(title)
                     .description(description)
                     .build();
-            tasksService.save(task).call();
+            tasksService.save(task);
```
Dorvaryn: We don't need Action as return value for service actions anymore since the service handles concurrency.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/task/data/model/Id.java
```diff
@@ -3,7 +3,7 @@
 import com.google.auto.value.AutoValue;
 
 @AutoValue
-public abstract class Id {
+public abstract class Id implements Comparable<Id> {
```
Dorvaryn: The data is ordered as a mean to ensure consistency during concurrent modification.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
```diff
@@ -82,24 +82,16 @@ public void call(Subscriber<? super SyncedData<Task>> subscriber) {
...
-    }
-
     private <T> Observable.Transformer<T, T> delay() {
         return new Observable.Transformer<T, T>() {
             @Override
             public Observable<T> call(Observable<T> observable) {
                 return observable.doOnEach(new Action1<Notification<? super T>>() {
                     @Override
                     public void call(Notification<? super T> t) {
+                        if (t.getKind() == Notification.Kind.OnCompleted) {
```
Dorvaryn: This is closed to actual API call behaviour, once we get a response onNext and onComplete are close together not delayed.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
```diff
@@ -48,18 +50,36 @@ public PersistedTasksService(
...
+                });
+    }
+
+    private void initialiseRelay() {
+        long actionTimeInMillis = clock.timeInMillis();
+        localDataSource.getTasks()
+                .flatMap(fetchFromRemoteIfOutOfDate(actionTimeInMillis))
+                .switchIfEmpty(fetchFromRemote(actionTimeInMillis))
+                .compose(asValidatedEvent(taskRelay.getValue()))
+                .subscribeOn(scheduler)
```
Dorvaryn: Scheduling is done by the service now.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
```diff
@@ -170,45 +174,53 @@ public Tasks call(List<Task> tasks) {
...
+                .subscribe();
+    }
+
+    private Func1<Tasks, Tasks> updateOnlyMostRecent(final long actionTimestamp) {
+        return new Func1<Tasks, Tasks>() {
+            @Override
+            public Tasks call(Tasks tasks) {
+                Tasks currentTasks = taskRelay.getValue().data().or(Tasks.empty());
+                return tasks.overrideWithMostRecentActionsFrom(currentTasks, actionTimestamp);
```
Dorvaryn: This ensures we don't keep local changes if they are more up to date.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
```diff
@@ -121,12 +129,22 @@ public boolean apply(Map.Entry<Id, SyncedData<Task>> input) {
         };
     }
 
+    public Tasks overrideWithMostRecentActionsFrom(Tasks tasks, final long actionTimestamp) {
+        ImmutableMapWithCopy<Id, SyncedData<Task>> filter = tasks.internalMap().filter(new Predicate<Map.Entry<Id, SyncedData<Task>>>() {
```
Mecharyry: Extract filter?

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
```diff
@@ -22,12 +22,12 @@
     private Tasks localTasks = Tasks.asSynced(
             ImmutableList.<Task>of(
                     Task.builder()
-                        .id(Id.from("42"))
```
Mecharyry: Was this value changed for any particular reason? Or just don't like the meaning of "life, the universe and everything?" 👅

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
Mecharyry: Sorry @Dorvaryn I don't understand what you mean. Are you saying this is close to the actual API call behaviour  in that it will return `onNext` and `onComplete` at almost the same time rather than having an `onNext` and then delaying the emission of the `onComplete`?

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
```diff
@@ -82,24 +82,16 @@ public void call(Subscriber<? super SyncedData<Task>> subscriber) {
...
-        return Observable.create(new Observable.OnSubscribe<Void>() {
-            @Override
-            public void call(Subscriber<? super Void> subscriber) {
-                localTasks = Tasks.empty();
-                subscriber.onCompleted();
-            }
-        }).compose(this.<Void>delay());
-    }
-
     private <T> Observable.Transformer<T, T> delay() {
```
Mecharyry: Used this in a couple of places, worth extracting?

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
```diff
@@ -101,46 +121,30 @@ public Tasks apply(Tasks input) {
...
-        });
-    }
-
-    private Func1<Tasks, Observable<Tasks>> fetchFromRemoteIfOutOfDate() {
+    private Func1<Tasks, Observable<Tasks>> fetchFromRemoteIfOutOfDate(final long actionTimeInMillis) {
         return new Func1<Tasks, Observable<Tasks>>() {
             @Override
             public Observable<Tasks> call(Tasks tasks) {
                 if (tasksDataFreshnessChecker.isFresh(tasks)) {
                     return Observable.just(tasks);
```
Mecharyry: Just for my curiosity, could you not have a filter on `localDataSource.getTasks()` instead of using an Observable.just? I guess you would have an extra couple of actions to perform in your pipeline.

Architecture/TodoApp/android/core/src/test/kotlin/com/novoda/todoapp/tasks/service/PersistedTasksServiceTest.kt
```diff
@@ -81,15 +86,18 @@ class PersistedTasksServiceTest {
...
         val localTasks = sampleLocalTasks()
         putTaskListIntoRemoteDataSource(remoteTasks)
         putTasksIntoLocalDataSource(localTasks)
         `when`(freshnessChecker.isFresh(localTasks)).thenReturn(true)
 
         val testObserver = subscribeToTasksEvent()
+        testScheduler.triggerActions()
 
         testObserver.assertReceivedOnNext(listOf(
+                emptyIdleEvent(),
```
Mecharyry: Why does this now return an `Idle` event first?

Architecture/TodoApp/android/core/src/test/kotlin/com/novoda/todoapp/tasks/service/PersistedTasksServiceTest.kt
```diff
@@ -852,36 +909,109 @@ class PersistedTasksServiceTest {
...
-                loadingEventWith(tasksWithCompletedMarkedDeleted()),
+                loadingEventWith(asSyncedTasks(tasks)),
+                loadingEventWith(tasksWithCompletedMarkedDeleted(TEST_TIME)),
                 loadingEventWith(asSyncedTasks(remoteTasksWithCompletedTasksRemoved())),
                 idleEventWith(asSyncedTasks(remoteTasksWithCompletedTasksRemoved()))
         ))
     }
 
     @Test
+    fun given_WeHavePendingTaskInTheService_on_ClearCompletedTasks_it_ShouldNotUpdatePendingTask() {
```
Mecharyry: Is there any way we can make these couple of tests a little bit more readable, maybe just with a given?

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
Dorvaryn: Yes, exactly in a real API call there would be next to no delay between the onNext and onComplete.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
Dorvaryn: It was just moved down, it is a switch since now items are ordered I found it easier to see them in the same order here. Not strictly needed.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/source/InMemoryLocalTaskDataSource.java
Dorvaryn: Only used in 2 places atm. Rule of 3 :trollface:

Architecture/TodoApp/android/core/src/test/kotlin/com/novoda/todoapp/tasks/service/PersistedTasksServiceTest.kt
Dorvaryn: With the testScheduler it will get this item since before it evens performs any action the original value is sent. Before since it was running the observable instantly it was not sent.

This is not a change in behaviour, just in the tests, we now get the full behaviour here to test => better tests.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
Dorvaryn: If I have a filter, how do I call to the remote ? Not sure what flow you are talking about. Can you draft an example ?

Mecharyry: Could you do:

```
        localDataSource.getTasks()
                .filter(new Func1<List<Integer>, Boolean>() {
                    @Override
                    public Boolean call(List<Integer> integers) {
                        return tasksDataFreshnessChecker.isFresh(tasks);
                    }
                })
                .switchIfEmpty(fetchFromRemote(actionTimeInMillis))
                .compose(asValidatedEvent(taskRelay.getValue()))
                .subscribeOn(scheduler)
                .subscribe(taskRelay);
```
Or have I missed something? 👅

Architecture/TodoApp/android/core/src/test/kotlin/com/novoda/todoapp/tasks/service/PersistedTasksServiceTest.kt
Mecharyry: ok, thanks for the explanation 👍

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
```diff
@@ -53,13 +53,21 @@ public boolean isEmpty() {
...
     }
 
-    public boolean hasNoActionMoreRecentThan(SyncedData<Task> syncedData) {
-        return hasNoActionMoreRecentThan(syncedData.data().id(), syncedData.lastSyncAction());
+    public boolean actionIsAbsentOrNoMoreRecentThan(SyncedData<Task> syncedData) {
+        return actionIsAbsentOrNoMoreRecentThan(syncedData.data().id(), syncedData.lastSyncAction());
     }
 
-    public boolean hasNoActionMoreRecentThan(Id id, long lastSyncAction) {
+    public boolean actionIsAbsentOrNoMoreRecentThan(Id id, long lastSyncAction) {
```
joetimmins: `actionIsAbsentOrOlderThan`

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
```diff
@@ -53,13 +53,21 @@ public boolean isEmpty() {
...
         SyncedData<Task> internalData = internalMap().get(id);
-        return internalData == null || lastSyncAction >= internalData.lastSyncAction();
+        return internalData == null || internalData.lastSyncAction() <= lastSyncAction;
+    }
+
+    public boolean actionExistsAndIsOutdatedComparedTo(SyncedData<Task> syncedData) {
```
joetimmins: `actionExistsAndIsOlderThan` ?

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
joetimmins: `filter` is a bad variable name here. How about `moreRecentActions` ? That way, line 139 will read `return Tasks.from(internalMap().putAll(moreRecentActions));` which makes it clearer that this `Tasks` object has been updated with the more recently updated `Task` objects from the parameter.

Architecture/TodoApp/android/core/src/test/kotlin/com/novoda/todoapp/tasks/service/PersistedTasksServiceTest.kt
```diff
@@ -852,36 +897,101 @@ class PersistedTasksServiceTest {
...
+                loadingEventWith(tasks),
+                loadingEventWith(tasksWithDeletedLocallyAndAheadData),
+                loadingEventWith(untouchedAheadData),
+                idleEventWith(untouchedAheadData)
+        ))
+    }
+
+    @Test
+    fun given_WeHaveCompletedPendingTaskInTheService_on_ClearCompletedTasks_it_ShouldNotUpdatePendingTask() {
+        val (tasks, task) = taskListWithOneAhead(taskListWithSomeCompleted(), 0)
```
joetimmins: Nice 👍

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
Dorvaryn: That solution will not send the outdated data to the relay. The current solution either sends the fresh data or will send the outdated one while fetching from remote an then update the relay again.

Architecture/TodoApp/android/core/src/test/kotlin/com/novoda/todoapp/tasks/service/PersistedTasksServiceTest.kt
Dorvaryn: Gotta love Kotlin ^^

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
Dorvaryn: I agree we need to change those names, however we need to be able to show that the different is not just in the exists vs absent. They are also checking for lowerThan vs LowerThanOrEquals.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
Mecharyry: ah ok, my bad xD

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
```diff
@@ -53,13 +53,21 @@ public boolean isEmpty() {
...
+        return actionExistsAndIsOutdatedComparedTo(syncedData.data().id(), syncedData.lastSyncAction());
+    }
+
+    public boolean actionExistsAndIsOutdatedComparedTo(Id id, long lastSyncAction) {
```
danybony: this could be private?

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
```diff
@@ -121,13 +117,27 @@ public boolean apply(Map.Entry<Id, SyncedData<Task>> input) {
...
+        return Tasks.from(internalMap().filter(new Predicate<Map.Entry<Id, SyncedData<Task>>>() {
+            @Override
+            public boolean apply(Map.Entry<Id, SyncedData<Task>> input) {
+                return predicate.apply(input.getValue());
+            }
+        }));
+    }
+
+    public Tasks insertOrUpdate(Tasks tasks) {
+        return Tasks.from(internalMap().putAll(tasks.internalMap()));
```
frapontillo: Maybe extract the `internalMap().putAll...` to a `newMap` var?

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
```diff
@@ -170,45 +175,67 @@ public Tasks call(List<Task> tasks) {
...
-            final Tasks allLocalTasks,
-            final Tasks activeLocalTasks,
-            final long actionTimestamp
-    ) {
+    private static boolean actionExistsAndIsOutdatedComparedTo(Tasks tasks, SyncedData<Task> syncedData) {
+        Optional<SyncedData<Task>> existingAction = tasks.get(syncedData.data().id());
+        return existingAction.isPresent() && existingAction.get().lastSyncAction() < syncedData.lastSyncAction();
+    }
+
+    private static Observable.Transformer<List<Task>, Tasks> confirmLocalDeletionsOrRevert(final BehaviorRelay<Event<Tasks>> taskRelay, final long actionTimestamp) {
```
frapontillo: This should go right after `clearCompletedTasks`.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
```diff
@@ -170,45 +175,67 @@ public Tasks call(List<Task> tasks) {
...
+        return new Func1<Tasks, Tasks>() {
             @Override
-            public void call() {
-                long actionTimestamp = clock.timeInMillis();
-                Event<Tasks> currentState = taskRelay.getValue();
-                Tasks allLocalTasks = currentState.data().or(Tasks.empty());
-                Tasks uncompletedLocalTasks = mapFunctionToTasks(allLocalTasks, markCompletedAsDeleted());
+            public Tasks call(Tasks tasks) {
+                Tasks currentTasks = taskRelay.getValue().data().or(Tasks.empty());
+                return tasks.insertOrUpdate(currentTasks.filter(isMoreRecentAction(tasks, actionTimestamp)));
```
frapontillo: I would really extract this to a `mostRecentTasks` variable and then use it to update `tasks`.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
```diff
@@ -254,75 +283,51 @@ private static Tasks mapFunctionToTasks(Tasks tasks, Function<SyncedData<Task>,
...
-                remoteDataSource.saveTask(updatedTask)
-                        .compose(startAheadThenConfirmOrMarkAsError(updatedTask, actionTimestamp))
-                        .compose(updateAndPersistIfMostRecentAction())
-                        .subscribe();
-            }
-        };
+    public void save(final Task updatedTask) {
+        final long actionTimestamp = clock.timeInMillis();
+        remoteDataSource.saveTask(updatedTask)
+                .compose(startAheadThenConfirmOrMarkAsError(updatedTask, actionTimestamp))
```
frapontillo: Please move `startAheadThenConfirmOrMarkAsError` after this method so it is more readable.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/service/PersistedTasksService.java
```diff
@@ -341,33 +346,46 @@ public void delete(final Task task) {
...
             }
 
             @Override
-            public Observable<Event<Tasks>> thenMap(Event<Tasks> value) {
-                return persistTasksEventLocally().call(value);
+            public Observable<Tasks> thenMap(final Event<Tasks> tasksEvent) {
+                Tasks tasks = tasksEvent.data().or(Tasks.empty());
+                return Observable.just(tasksEvent.updateData(tasks))
+                        .doOnNext(taskRelay)
+                        .flatMap(new Func1<Event<Tasks>, Observable<Tasks>>() {
```
takecare: couldn't the `Func1` be extracted to a method?

Architecture/TodoApp/android/core/src/test/kotlin/com/novoda/todoapp/tasks/service/PersistedTasksServiceTest.kt
```diff
@@ -1081,15 +1200,22 @@ class PersistedTasksServiceTest {
...
+                SyncedData.from(Task.builder().id(Id.from("425")).title("Whizz").build(), SyncState.IN_SYNC, TEST_TIME)
         )))
     }
 
     private fun asSyncedTasks(remoteTasks: List<Task>) = Tasks.asSynced(remoteTasks, TEST_TIME)
 
     private fun noEmptyTasks() = NoEmptyTasksPredicate()
 
+
+    private fun givenPresenterWithTestScheduler(): TestScheduler {
```
frapontillo: Reading this gives me the idea that it returns a presenter. If you still want to return a `TestScheduler`, instead of setting it as an object property, you can refactor it to `givenPresenterAndReturnTestScheduler` or smt like that, maybe?

Architecture/TodoApp/android/mobile/src/main/java/com/novoda/todoapp/tasks/service/TasksServiceObserveOn.java
```diff
@@ -0,0 +1,76 @@
...
+
+import rx.Observable;
+import rx.Scheduler;
+
+public class TasksServiceObserveOn implements TasksService {
+
+    private final TasksService tasksService;
+    private final Scheduler scheduler;
+
+    public TasksServiceObserveOn(TasksService tasksService, Scheduler scheduler) {
```
frapontillo: 👍

frapontillo:
This PR has sweat, blood and tears all over it: excellent job 💯 👏 🎆 ✨ !
It just needs some love wrt/ moving methods as soon as they're used, since I found it hard going up and down between 900 LoC. I marked some of them, but the same applies to methods used in the methods to be moved as well.

Architecture/TodoApp/android/core/src/main/java/com/novoda/todoapp/tasks/data/model/Tasks.java
Dorvaryn: Not sure it adds much here.

Dorvaryn:
There will be a clean up of the service class, it is too big now but I want to do this in another PR.

frapontillo:
Final LGTM 👍 :shipit:

PR #144, opened 2016-08-01
Add monkey runner plugin

ataulm:
Problem: we were copy/pasting some of these classes between projects. It allows us to start the monkey trap before running the android monkey. The monkey trap is another app that draws over the system notification bar so the monkey cannot open it (in theory... cough cough samsung).

Solution: we extract a plugin so it can be used between projects.

Paired with:
@danybony , @mr-archano , @xrigau


monkey-runner/build.gradle
```diff
@@ -0,0 +1,17 @@
+buildscript {
+    repositories {
+        jcenter()
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:2.1.2'
```
mr-archano: It would be good to pull out all these dependencies in a separate file in order to easy their maintenance.

monkey-runner/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Tue Oct 20 10:41:37 BST 2015
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.13-all.zip
```
mr-archano: Latest version is [2.14.1](https://docs.gradle.org/current/release-notes). Have you tried using it?

monkey-runner/plugin/build.gradle
```diff
@@ -0,0 +1,20 @@
+apply plugin: 'com.novoda.bintray-release'
+apply plugin: 'groovy'
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile gradleApi()
+    compile 'com.novoda:gradle-android-command-plugin:1.6.2'
```
mr-archano: The other build.gradle is using `1.5.0`. Using common declared dependencies could help avoiding such issues.

monkey-runner/sample/buildSrc/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+repositories {
+    jcenter()
+}
+
+// these must mirror the ones in the `plugin` module
+// the dependencies are not transitive (we specify only sourceset below)
+dependencies {
+    compile gradleApi()
+    compile 'com.novoda:gradle-android-command-plugin:1.6.2'
+}
```
mr-archano: you could try to factor out this common setup in a `shared.gradle` that could be applied both here and in `plugin/build.gradle`

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
```diff
@@ -0,0 +1,96 @@
+package com.novoda.monkey
+
+import org.gradle.api.GradleException
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+
+public class MonkeyConfigurationPlugin implements Plugin<Project> {
+
+    private static final String EXTENSION_NAME = 'monkeyRunner'
```
juankysoriano: I think `MONKEY_RUNNER_EXTENSION` would get more readable, the same for the other constants, let's be more explicit

juankysoriano:
Not much to add apart of something about the constants naming

xrigau:
The API might be a bit limited but this is a good starting point 👍

ataulm:
Just extracting the dependencies

BuildPropertiesPlugin/build.gradle
```diff
@@ -2,4 +2,12 @@ ext {
   pluginGroup = 'com.novoda'
   pluginVersion = '1.0.1'
   pluginId = 'build-properties-plugin'
+
+  buildDependencies = [
```
ataulm: nao

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
ataulm: renamed here - https://github.com/novoda/spikes/pull/144/commits/bccb50da5cdaf27aa5769d20c6fcd2732818c722

monkey-runner/gradle/wrapper/gradle-wrapper.properties
ataulm: Updated to the latest version and it seems there are no issues. I added the wrapper task so that I could specify the version in the `dependencies.gradle` once rather than in each place separately:

https://github.com/novoda/spikes/pull/144/commits/f09336b4fffdbc929d6fe311b1cc07a13afe4566

monkey-runner/build.gradle
ataulm: pulling put the build script dependencies - https://github.com/novoda/spikes/pull/144/commits/109e83666a9ea1864b5c3813634fa7410ac2fa76

pulling out the compile dependencies - https://github.com/novoda/spikes/pull/144/commits/722d9693d24b9d9a1edc2b205c95386caebeb170

monkey-runner/sample/buildSrc/build.gradle
ataulm: I avoided doing this so the dependencies closure was visible in each file - I tried to extract just the contents of the block but it didn't like it.

The other change (extracting the dependencies to a single file) should address half the beef with this comment, at least (don't have keep the gradle android command plugin version synced in two places).

ataulm:
:shipit: if it's good @mr-archano :D

monkey-runner/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+        jcenter()
+    }
+    dependencies {
+        classpath gradlePlugins.gradle
+        classpath gradlePlugins.gradleAndroidCommandPlugin
+        classpath gradlePlugins.bintrayRelease
+    }
+}
+
+apply from: 'dependencies.gradle'
```
mr-archano: if you apply at this level you can save the one above in  the`buildscript` closure, just move it above that block

monkey-runner/dependencies.gradle
```diff
@@ -0,0 +1,18 @@
+ext {
```
mr-archano: ✨

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
```diff
@@ -0,0 +1,96 @@
...
+        void ensureMandatoryPropertiesPresent() {
+            if (taskDependency == null) {
+                notifyMissingProperty('taskDependency')
+            }
+            if (packageNameFilter == null) {
+                notifyMissingProperty('packageNameFilter')
+            }
+        }
+
+        static void notifyMissingProperty(String propertyName) {
```
mr-archano: `private`?

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/TargetedMonkey.groovy
```diff
@@ -0,0 +1,24 @@
...
+
+import com.novoda.gradle.command.Monkey
+
+class TargetedMonkey extends Monkey {
+
+    def packageName
+    def logFileName
+
+    @Override
+    def getPackageName() {
```
mr-archano: given `packageName` is not private you get public [getter and setter](http://groovy-lang.org/style-guide.html#_getters_and_setters) for free, therefore you don't need to re-define the getter here

monkey-runner/sample/build.gradle
```diff
@@ -0,0 +1,22 @@
+task wrapper(type: Wrapper) {
+    gradleVersion = versions.gradleWrapperVersion
+}
+
+buildscript {
+    apply from: '../dependencies.gradle'
```
mr-archano: same as [the comment above](https://github.com/novoda/spikes/pull/144/files#r73426394)

monkey-runner/build.gradle
ataulm:
```groovy
apply from: 'dependencies.gradle'
buildscript {
    repositories {
        jcenter()
    }
    dependencies {
        classpath gradlePlugins.gradle
        classpath gradlePlugins.gradleAndroidCommandPlugin
        classpath gradlePlugins.bintrayRelease
    }
}
allprojects {
    repositories {
        jcenter()
    }
    version = '0.0.1'
}
```

This didn't work - "Could not get unknown property 'gradlePlugins' for object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler".

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/TargetedMonkey.groovy
ataulm: 👍  https://github.com/novoda/spikes/pull/144/commits/01f7dfd31ab9e7994745831d64a5e2d6681ccd0d

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
ataulm: 👍  fd85c09

monkey-runner/build.gradle
mr-archano: I suppose this doesn't work because you're defining the buildscript repos and the dependencies for the root project, while you actually need them only in the subprojects. Try with:
```
apply from: 'dependencies.gradle'
subprojects {
    buildscript {
        repositories {
            jcenter()
        }
        dependencies {
            classpath gradlePlugins.gradle
            classpath gradlePlugins.gradleAndroidCommandPlugin
            classpath gradlePlugins.bintrayRelease
        }
    }
    repositories {
        jcenter()
    }
    version = '0.0.1'
}
```

xrigau: I think the problem is you can only define the `buildscript` closure once so the plugin shouldn't modify that https://discuss.gradle.org/t/how-can-i-modify-the-buildscript-blocks-dependencies-i-e-classpath-from-within-a-custom-plugin/5612/2

mr-archano: @xrigau I actually believe that is a related but slightly different issue, that can be resolved by introducing `sample/bulidSrc` module for instance.
The reason why apply here doesn't work is because the `buildscript` configuration of the root project configuration is evaluated while the properties extension is not yet part of the model, therefore the error.

ataulm: Sorry for the delay :o - fixed here 85fe33d.

As the task wrapper was using a version defined in dependencies.gradle, restricting it to subprojects meant that the task wrapper failed, but instead of having a separate file to define the wrapper version, I opted for the duplication in `/build.gradle` and `/sample/build.gradle` with regards to the declaration of the wrapper version.

xrigau:
LGTM 👍  hopefully @mr-archano has no more beef with this, otherwise can be addressed in another PR - LET'S :shipit: !

xrigau:
btw I found your easter egg @ataulm 😂 

screen shot 2016-08-08 at 12 50 17 pm

ataulm:
eagle eyes!

PR #146, opened 2016-08-17
BuildPropertiesPlugin: Remove whole properties support

mr-archano:
While planning to add fallback support to the plugin I realised `buildConfigProperties` / `resValueProperties` should not be part of the core features, because not easy to integrate with the upcoming facilities. 
The logic used to generate fields and resource names from a whole properties file could be easily scripted separately using the facilities provided by the plugin.


BuildPropertiesPlugin/plugin/src/main/groovy/com/novoda/buildproperties/Entry.groovy
```diff
@@ -0,0 +1,37 @@
...
+    Double getDouble() {
+        getValue() as Double
+    }
+
+    String getString() {
+        getValue() as String
+    }
+
+    Object getValue() {
+        value.call()
```
frapontillo: In a future `or` implementation, this is where you'd want to catch something like `CompositeException` and, if `value.call()` fails, you add the new exception to the previously thrown `CompositeException`.
This way you have an exception that says something like 'Could not find values for keys: a, b, c, d.'.

frapontillo:
Does this PR make the plugin oriented to any Java module rather than Android specifically? Is there anything you can remove to achieve that?

mr-archano:
@frapontillo the plugin internal facilities don't know anything about Android already. At the moment we only augment Android specific facilities (like buildConfig, resValue and signingConfig), hence why we don't support plain java projects. It would be nice to understand what we want to achieve there, maybe another RFC would help.

PR #149, opened 2016-08-23
[Slacker] stackoverflow random question

takecare:
Picks a random unanswered question and displays it.

screen shot 2016-08-23 at 13 30 02

### DIFF
--- a/slacker/public/widgets/stackoverflow.js
+++ b/slacker/public/widgets/stackoverflow.js
@@ -15,21 +15,21 @@ function presentQuestion(data) {
     '<span class="active-channel-header">' + '&#128172;' + '</span>' +
     '<br>' +
     '<div>' +
-      '<span class="active-channel-copy">We have ' + linkifyCount(data.questions) + '!</span>' +
-      '<p class="">Here\'s one: ' + linkifyQuestion(pickRandomQuestion(data.questions)) + '</p>' +
+      '<span class="active-channel-copy">We have ' + linkifyCount(data) + '!</span>' +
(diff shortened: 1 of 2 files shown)


blundell:
but but but nobody can click those links. Why don't you display an surl.novoda that they can go to?

takecare:
i guess that can be done. do we have `surl`s for this already? also, i figured this can be clickable in some situations (e.g. you're at home and want to have a broad look on company stuff; the dashboard is your "new tab" default page -- this can be a chrome extension)

blundell:
this is how you can add urls: http://surl.novoda.com/content.py

Making this a chrome extension is a scary use case that I wouldn't want to make a priority.
For instance its completely fine for you to do whatever you want with the dashboard i.e. adding these types of things and making a chrome extension.
However the big blue URL looks really wrong on a dashboard and wouldn't work shown up around our offices on the wall displays. So you could consider not doing it / putting it behind a feature flag / config option / I don't know other idea. :)

takecare:
having a config option would be a good idea, i.e. running locally vs running as a wall-display dashboard.

Electryc:
Let's go retro - QR code for the wall-display 😸

takecare:
@blundell 
screen shot 2016-08-30 at 10 07 33

blundell:
nice!

blundell:
LGTM i can ship unless you waiting on moar?

takecare:
you can ship!

friedger:
Use nearby and physical Web

Am 30.08.2016 11:38 schrieb "Paul Blundell" <notifications@github.com>:

PR #150, opened 2016-08-25
BuildPropertiesPlugin: Introduce fallback support

mr-archano:
Every subclass of `Entries` (eg: `FilePropertiesEntries`) defines if the evaluation of an `Entry` will throw when a property is missing. In such cases we may want to avoid to break the build and to define a fallback via the `or()` operator instead. A fallback can be:

- another `Entry`, eg: `entry.or(buildProperties.secrets['fallback'])`
- a `Closure`, eg: `entry.or({ Math.random() })`
- a simple value, eg: `entry.or('fallback')`

(where `entry` is an instance of `Entry`, eg: `buildProperties.secrets['foo']`)

If the whole fallback chain fails then a `CompositeException` is thrown listing all
the causes of failure encountered evaluating the chain, eg:

```
A problem occurred while evaluating entry:
- No value defined for property 'foo' in 'secrets' properties
- No value defined for property 'fallback' in 'other' properties
```

### DIFF
--- a/BuildPropertiesPlugin/plugin/build.gradle
+++ b/BuildPropertiesPlugin/plugin/build.gradle
@@ -24,6 +24,8 @@ dependencies {
   testCompile 'com.google.guava:guava:19.0'
   testCompile 'com.google.truth:truth:0.28'
   testCompile 'com.android.tools.build:gradle:2.1.2'
+  testCompile 'com.google.code.findbugs:jsr305:3.0.0'
+
 }
 
 // https://code.google.com/p/android/issues/detail?id=64887
--- a/BuildPropertiesPlugin/plugin/src/main/groovy/com/novoda/buildproperties/Entry.groovy
+++ b/BuildPropertiesPlugin/plugin/src/main/groovy/com/novoda/buildproperties/Entry.groovy
@@ -34,4 +34,29 @@ class Entry {
         value.call()
     }
 
+    Entry or(def fallback) {
+        def other = from(fallback)
+        new Entry(key, {
+            try {
+                return getValue()
+            } catch (Throwable e) {
+                try {
+                    return other.call()
+                } catch (Throwable e2) {
+                    throw CompositeException.from(e).add(e2)
+                }
+            }
+        })
+    }
+
+    private static def from(def fallback) {
+        if (fallback instanceof Entry) {
+            return { (fallback as Entry).getValue() }
+        }
+        if (fallback instanceof Closure) {
+            return fallback
+        }
+        return { fallback }
+    }
+
 }
--- a/BuildPropertiesPlugin/plugin/src/test/groovy/com/novoda/buildproperties/CompositeExceptionSubject.groovy
+++ b/BuildPropertiesPlugin/plugin/src/test/groovy/com/novoda/buildproperties/CompositeExceptionSubject.groovy
@@ -9,13 +9,13 @@ import javax.annotation.Nullable
 
 final class CompositeExceptionSubject extends Subject<CompositeExceptionSubject, CompositeException> {
 
-    private static
-    final SubjectFactory<CompositeExceptionSubject, CompositeException> FACTORY = new SubjectFactory<CompositeExceptionSubject, CompositeException>() {
-        @Override
-        CompositeExceptionSubject getSubject(FailureStrategy fs, CompositeException that) {
-            new CompositeExceptionSubject(fs, that)
-        }
-    }
+    private static final SubjectFactory<CompositeExceptionSubject, CompositeException> FACTORY =
+            new SubjectFactory<CompositeExceptionSubject, CompositeException>() {
+                @Override
+                CompositeExceptionSubject getSubject(FailureStrategy fs, CompositeException that) {
+                    new CompositeExceptionSubject(fs, that)
+                }
+            }
 
     public static CompositeExceptionSubject assertThat(CompositeException compositeException) {
         Truth.assertAbout(FACTORY).that(compositeException)
@@ -24,10 +24,6 @@ final class CompositeExceptionSubject extends Subject<CompositeExceptionSubject,
     private CompositeExceptionSubject(FailureStrategy failureStrategy,
                                       @Nullable CompositeException subject) {
         super(failureStrategy, subject)
-    }
-
-    public void contains(Throwable... throwables) {
-        Truth.assertThat(subject.exceptions).containsExactly(throwables)
     }
 
     public void hasMessage(String... messages) {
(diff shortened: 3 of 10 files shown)


BuildPropertiesPlugin/plugin/build.gradle
```diff
@@ -24,6 +24,8 @@ dependencies {
   testCompile 'com.google.guava:guava:19.0'
   testCompile 'com.google.truth:truth:0.28'
   testCompile 'com.android.tools.build:gradle:2.1.2'
+  testCompile 'com.google.code.findbugs:jsr305:3.0.0'
```
blundell: 🎉

BuildPropertiesPlugin/plugin/src/test/groovy/com/novoda/buildproperties/CompositeExceptionSubject.groovy
```diff
@@ -0,0 +1,37 @@
...
+import com.google.common.truth.FailureStrategy
+import com.google.common.truth.Subject
+import com.google.common.truth.SubjectFactory
+import com.google.common.truth.Truth
+
+import javax.annotation.Nullable
+
+final class CompositeExceptionSubject extends Subject<CompositeExceptionSubject, CompositeException> {
+
+    private static
```
blundell: orlly

BuildPropertiesPlugin/plugin/src/test/groovy/com/novoda/buildproperties/EntrySubject.groovy
```diff
@@ -0,0 +1,49 @@
...
+        Truth.assertAbout(FACTORY).that(entry)
+    }
+
+    private EntrySubject(FailureStrategy failureStrategy, @Nullable Entry subject) {
+        super(failureStrategy, subject)
+    }
+
+    public void willThrow(Class<? extends Throwable> throwableClass) {
+        try {
+            subject.value
```
blundell: this is groovy? So when you attempt to access value and it's not there it throws?

BuildPropertiesPlugin/plugin/src/test/groovy/com/novoda/buildproperties/CompositeExceptionSubject.groovy
mr-archano: auto-format fail

BuildPropertiesPlugin/plugin/src/test/groovy/com/novoda/buildproperties/CompositeExceptionSubject.groovy
```diff
@@ -0,0 +1,33 @@
...
+    public static CompositeExceptionSubject assertThat(CompositeException compositeException) {
+        Truth.assertAbout(FACTORY).that(compositeException)
+    }
+
+    private CompositeExceptionSubject(FailureStrategy failureStrategy,
+                                      @Nullable CompositeException subject) {
+        super(failureStrategy, subject)
+    }
+
+    public void hasMessage(String... messages) {
```
frapontillo: 💯

BuildPropertiesPlugin/plugin/src/test/groovy/com/novoda/buildproperties/EntrySubject.groovy
mr-archano: Yes. This line is equivalent to `getSubject().getValue()` that in turns [evaluates the closure](https://github.com/novoda/spikes/blob/master/BuildPropertiesPlugin/plugin/src/main/groovy/com/novoda/buildproperties/Entry.groovy#L33-L35) inside the `Entry`. At the moment the only implementation available [will throw an `IllegalArgumentException`](https://github.com/novoda/spikes/blob/master/BuildPropertiesPlugin/plugin/src/main/groovy/com/novoda/buildproperties/FilePropertiesEntries.groovy#L48) when the specified property is not found.

BuildPropertiesPlugin/plugin/src/test/groovy/com/novoda/buildproperties/EntryTest.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+    private static final RuntimeException EXCEPTION_1 = new RuntimeException("exception 1")
+    private static final RuntimeException EXCEPTION_2 = new RuntimeException("exception 2")
+
+    @Test
+    public void shouldReturnFirstEntryValueInOrWhenFirstEntryValueDoesNotThrow() {
+        Entry entry1 = new Entry('key1', { 'value1' })
+        Entry entry2 = new Entry('key2', { 'value2' })
+
+        Entry or = entry1.or(entry2)
```
frapontillo: `Entry entry`?

BuildPropertiesPlugin/plugin/src/main/groovy/com/novoda/buildproperties/Entry.groovy
```diff
@@ -34,4 +34,29 @@ class Entry {
...
+    Entry or(def fallback) {
+        def other = from(fallback)
+        new Entry(key, {
+            try {
+                return getValue()
+            } catch (Throwable e) {
+                try {
+                    return other.call()
+                } catch (Throwable e2) {
+                    throw CompositeException.from(e).add(e2)
```
frapontillo: I was thinking if there were a way to, besides give a good description of the exceptions, nest the causes of the exceptions so it's clear from the stacktrace that if the fallback has failed was because the default value closure failed as well.

This is clear from both the spec and the exception message, but it could be even more explicit via the stacktrace.

frapontillo: Yes, that would be too much effort. Isn't the cause implicit if you do `new Exception(cause)`?

frapontillo:
LGTM let's 🚢

BuildPropertiesPlugin/plugin/src/main/groovy/com/novoda/buildproperties/Entry.groovy
mr-archano: yes it is, but then you also need to clean up the stack trace of the wrapper exception to avoid showing misleading traces like 
```
at com.novoda.buildproperties.CompositeException.add(CompositeException.groovy:30)
	at com.novoda.buildproperties.CompositeException$add$0.call(Unknown Source)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:113)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:125)
```
As I said is a matter of amount effort needed vs actual gain.

PR #152, opened 2016-08-30
Slacker~

ouchadam:
A dashboard of dashboards with the intention of being incredibly easy to build upon.

NodeJS server with simple JS client.

The format is 

Client: Widget 

```javascript
if (widgets.ciWall) throw "CiWall already exists";

widgets.ciWall = function() {
  return {
    present: function(data, element) {
      element.innerHTML = '<iframe src="' + data.payload + '" class="frame"></iframe>';
    }
  };
}
```

Server: Widget Payload 

 ```javascript
module.exports = {
  rule: ciWall,
  rank: 1
}

function ciWall() {
  var result = function(resolve, reject) {
    resolve({
      widgetKey: 'ciWall',
      payload: 'https://ci.novoda.com//plugin/jenkinswalldisplay/walldisplay.html?viewName=Active&jenkinsUrl=https%3A%2F%2Fci.novoda.com%2F'
    });
  }
  return new Promise(result);
}
```

TODO 
- To be able to provide a `module` which contains both the server and client side components


frapontillo:
I kind of get the reasons behind this approach, but I would like to suggest a different implementation of the same thing: [Web Components](http://webcomponents.org/).

They can easily be polyfilled with Polymer for older browsers, and can be tested in isolation, having HTML (view) and JS (ViewModel or Controller or ViewWhatever).

This will also allow us to bundle those components in any Web page, regardless of the underlying "widget" system.

frapontillo:
Meanwhile, in Sahara...

PR #154, opened 2016-09-03
BPP: Add lazy-loading of properties files

mr-archano:
At the moment the project evaluation fails straight away when `buildProperties` contains a set of entries from a file that is not found.
It would be useful in multiple scenarios to lazy load properties file and fail the build only when one of their entries will actually be evaluated.


mr-archano:
@hal9002 test this please.

mr-archano:
@hal9002 retest this please.

mr-archano:
@hal9002 retest this please

PR #158, opened 2016-09-27
Tpbot/add readme

danybony:
Create the TelepresenceBot directory which will contain the project source (for both Android and Arduino) and adds the first draft of the readme.


Mecharyry:
I was looking for this mystery directory, turns out there is still a PR open! 👅

PR #159, opened 2016-09-27
Add initial project structure for tpbot

Mecharyry:
Add initial project structure for tpbot. Will build on this to create the sender and receiver for the tpbot.


danybony:
LGTM :+1:

PR #161, opened 2016-10-04
Use new app icon

alexstyl:
Add a new icon to the Android app

ic_launcher

what? It is important 😄


amlcurran:
What is tpbot? And have you got some assurance that the copyright is OK here? I'd imagine we can't use a image of Wall-e...

ouchadam:
+1 for avoiding any IPs

alexstyl:
Fair enough. I'll close this one then and open a new one when I make a proper icon :)

PR #164, opened 2016-10-12
Display the Set Live Wallpaper Settings when opening MainActivity

alexstyl:
Our current MainActivity displays nothing and does nothing.
This PR launches the Set Live wallpaper setting with our wallpaper selected.

setwallpaper

Eventually we could have something more meaningful such as a configuration screen or some about info, but that works for now 🙃


lgvalle:
💞

PR #171, opened 2016-10-18
Load assets according to time of day

alexstyl:
In this PR we are actually switching out the displaying assets of the time of day we want to load. Previously we were loading 3 different layers at the same time. Now we are only loading one each time (as the images are static)

For the purpose of the screenshot I am switching out each asset everytime the screen is on:

| Switching in action |
| --- | 
| wallapper |

_it looks better in person..._ :(

### DIFF
--- a/FacesofLondon/app/src/main/java/com/novoda/wallpaper/droidcon/LondonParallaxWallpaperRenderer.java
+++ b/FacesofLondon/app/src/main/java/com/novoda/wallpaper/droidcon/LondonParallaxWallpaperRenderer.java
@@ -8,7 +8,9 @@ import javax.microedition.khronos.egl.EGLConfig;
 import javax.microedition.khronos.opengles.GL10;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import uk.co.halfninja.wallpaper.parallax.gl.Capabilities;
 import uk.co.halfninja.wallpaper.parallax.gl.Quad;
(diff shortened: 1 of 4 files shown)


FacesofLondon/app/src/main/java/com/novoda/wallpaper/droidcon/LondonParallaxWallpaperRenderer.java
```diff
@@ -162,6 +153,10 @@ void setOffset(float xOffset) {
     }
 
     void loadAssetsFor(TimeOfDay timeOfDay) {
-        // TODO for a follow up PR
+        try {
+            reloadLayersFor(timeOfDay);
+        } catch (IOException e) {
+            e.printStackTrace();
```
xrigau: would be nice to deal with this

FacesofLondon/app/src/main/java/com/novoda/wallpaper/droidcon/TimeOfDayCalculator.java
```diff
@@ -1,12 +1,18 @@
...
+
     TimeOfDay currentTimeOfDay() {
-        Calendar calendar = Calendar.getInstance();
-        int hourOfDay = calendar.get(Calendar.HOUR_OF_DAY);
-        return TimeOfDay.forHourOfDay(hourOfDay);
+        TimeOfDay[] values = TimeOfDay.values();
+        flag++;
+        if (flag >= values.length) {
+            flag = 0;
+        }
```
devisnik: Or `flags = (flags+1) % values.length` ?

alexstyl: This was actually some code I used for easy debugging and wasn't meant to be in the PR 😅 
Nice way of writing it though. Never occurred to me like that

FacesofLondon/app/src/main/java/com/novoda/wallpaper/droidcon/TimeOfDayCalculator.java
```diff
@@ -1,12 +1,18 @@
 package com.novoda.wallpaper.droidcon;
 
-import java.util.Calendar;
-
 class TimeOfDayCalculator {
```
alexstyl: I think you are right. Since `TimeOfDay.forHourOfDay(hourOfDay)` was moved into the same class, might as well move this there too

alexstyl: I wouldn't want the `TimeOfDay` itself to be dependent on `Calendar`. 

I find it perfectly fine for the `TimeOfDay` to know which parts of the day include which hours of the day, but it's the calendar that knows what time it is.

alexstyl:
Closing this as I found an issue. will open again with fix

PR #172, opened 2016-10-19
Gradle diff task dependencies demo

frapontillo:
### TL;DR

This PR introduces a future plugin to add conditional dependencies to tasks, according to the files that have been changed in the current PR.

### Objective

The aim is to enable mixed Android-iOS teams to share a single Github repository and to lessen the workload on the CI servers.
For instance, when something changes in the Android project directory we might want to execute only Android tests, since no code is shared with the iOS project.

This is essentially what is already happening (in a more hardcoded way) on the 10ft-tv repo with the [**`lightTest` task**](https://github.com/novoda/10ft-tv/blob/develop/tasks/light-test.gradle).

__________

The PR only contains a demo `build.gradle` to introduce the way to use the plugin.

The idea is to add extension methods to the `Task` class, so that it is able to conditionally call the "real" implementation of `dependsOn`.


gradle-diff-task-dependencies/demo/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+task('apple-test') << {
+    println 'this is an apple test task'
+}
+
+task('all-test') << {
+    println 'this is a test for everything'
+}
+
+task('selectiveTest')
+        .ifDiffMatches(/android\/.*/).dependsOn('android-test')
```
frapontillo: What I want to build. This PR is mostly oriented to the API: is this good or should I take a different approach?

PR #173, opened 2016-10-24
github to slack mergeability checker

ouchadam:
Nodejs app to check the open PRs of a github repository and ping slack if they're deemed unmergable

screenshot from 2016-10-24 16-34-21

paired ish~ with @ataulm


xrigau:
👍 would be more generic if the slack part was done in index.js instead of in mergeable.js

PR #175, opened 2016-11-02
Drop cap fixes

Mecharyry:
#### Problem
When editing the size of the DropCap and changing the length of the text input it is possible that the DropCap will overlap the copy text (below).

| | | |
|:----:|----|----|
| **Before** |drop_cap| |
| **After** | drop_cap | not_enough_text |

The second problem occurs when editing the text to display. I wrongly assumed that a call to `requestLayout` would perform a `measure` and `draw` call...which it does...but...apparently a `requestLayout` is always paired with a call to `invalidate`. 

#### Solution
The overlapping text was a simple case of ensuring that the calculation to determine how far the DropCap should span for is calculated for each measure where the `DropCapStaticLayout` is remeasured.

Whenever performing any task that requires a remeasure and a redraw we should have the pair of calls `requestLayout` and `invalidate`. When performing a task that requires just a redraw then we should only use `invalidate`.

### DIFF
--- a/drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
+++ b/drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
@@ -260,7 +260,12 @@ public class DropCapView extends View {
         int widthWithoutPadding = totalWidth - horizontalPadding;
 
         measureDropCapFor(widthWithoutPadding);
-        measureCopyFor(widthWithoutPadding);
+
+        if (enoughLinesForDropCap()) {
+            measureRemainingCopyFor(totalWidth);
+        } else {
+            measureWholeTextFor(totalWidth);
+        }
 
         int desiredHeight = dropCapLineHeight + copyStaticLayout.getHeight() + getPaddingTop() + getPaddingBottom();
         int desiredHeightMeasureSpec = MeasureSpec.makeMeasureSpec(desiredHeight, MeasureSpec.EXACTLY);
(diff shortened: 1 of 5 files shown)


drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
```diff
@@ -126,16 +129,18 @@ public void setCopyFontType(String fontPath) {
 
         hasPaintChanged = true;
         requestLayout();
+        invalidate();
     }
 
-    public void setDropCapTextSize(@DimenRes float textSize) {
+    public void setDropCapTextSize(float textSize) {
```
mr-archano: `float` should imply pixels. Providing `dp` would not make much sense because it will force you to evaluate the value in pixels later. Another option is to leverage resources instead and have `@DimenRes int textSizeResId` as parameter. But in that case maybe you could consider just an overload for `setDropCapTextSize`.

eduardb: But aren't pixels indivisible? Anyway, suffixing the argument with `Px` is enough for me.

drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
```diff
@@ -95,6 +95,8 @@ private void applyCustomAttributes(Context context, AttributeSet attrs) {
             copyTextPaint.setTextSize(copyTextSize);
             copyTextPaint.setColor(copyTextColor);
 
+            String text = typedArray.getString(R.styleable.DropCapView_text);
```
Mecharyry: That's a good question. Most likely I did it quickly without thinking, I'll update to use the android version, thanks 😃

drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
Mecharyry: Thanks for asking the question @eduardb as @mr-archano says it is in `px`, I'll update the parameter to be more explicit. I'll add the overload too, as you suggested @mr-archano 👍

Mecharyry:
After the reviews by @eduardb and @mr-archano I went back through some of the things in the `DropCapView`. Now introducing,  the NEW and IMPROVED `DropCapView` 👅 

- I've created a `setTextSize` for both the `DropCap` and the `Copy` for which the user can supply a unit to convert from.
- Removed the boolean flags for detecting when a new layout is required, decided to just null out both of the layouts in this eventuality.
- Encapsulated the null out, request layout and invalidation into a `remeasureAndRedraw` method, `re` because it will only trigger after the initial measure and draw of the view when it is inflated.
- Use the `setters` in the constructor rather than duplicating all but a few lines of code from the `setters` 
- Use the `android:text` attribute instead of a custom one for specifying the text.
- `MeasureCopyFor`was a large method that would either measure all the text or only the remaining text after the `DropCap` and its associated copy. This was split into two explanatory methods and the `if statement` moved to the measure. 

Hopefully, all of these changes will make the `DropCapView` more readable and ease maintenance in the future. Thanks guys for giving me the inspiration 😄

drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
```diff
@@ -6,15 +6,20 @@
...
 import android.text.Layout;
 import android.text.StaticLayout;
 import android.text.TextPaint;
 import android.util.AttributeSet;
+import android.util.TypedValue;
 import android.view.View;
 
 public class DropCapView extends View {
 
+    private static final int TEXT_SET_INDEX = 0;
```
eduardb: This is actually not needed. You can add `<attr name="android:text" />` in the `DropCapView` stylable, and reference it with `R.styleable.DropCapView_android_text`. You save a lot of code 👅

Mecharyry: Oh really?! Awesome! Thanks man 😄 learning more about attributes 💯

Mecharyry:
@eduardb This is one of the things that may need customising actually. Originally the idea was that if there is not enough copy text to wrap the DropCap then it will default to behaving like an ordinary  `TextView` which means using the same styles for all of the text. When there is enough text then it will use a custom styling for the DropCap. 

What do you think? Does it make sense the way I have done it or should it always use the specified styles?

eduardb:
I see. But is this behavior specified anywhere? I didn't know about it, and maybe other users of the library will be confused about it too. So either this should be specified in a highly visible place, so it's hard not to see it, but I would like it more if I am given the choice to have this behavior or to always show the DropCap, even for short texts.

Mecharyry:
@eduardb I've updated the readme to show the current behavior. I think changing the code to provide the choice will be a little bit of work, thinking about how it is currently implemented, so I will do that in a follow-up PR 😄 Thanks for the comments man, I really appreciate it.

eduardb:
You're welcome, anytime! :)

PR #179, opened 2016-11-03
Add widgets for Sonar, StackOverflow, App reviews and CI wall

xrigau:
We have done multiple things:

 - The dashboards now require the [slacker](https://github.com/novoda/spikes/tree/master/slacker) nodejs backend in order to work, we just have dummy clients that are used as the front-end.
 - The dashboard carousel communicates with the backend using websockets.
 - Rewrite the Sonar component to be just a dummy client instead of actually hitting the Sonar API.
 - The carousel contains a `DynamicComponent` that inflates each widget `Component` depending on the event received from the socket and we just show 1 widget at a time.
 - Added widgets for StackOverflow, Play store reviews and external URL (ci wall)
 - Simplified the addition of a new widget: As long as the backend sends the new widget event, you just have to add a new line to the `COMPONENTS` list in the `DashboardCarouselComponent` and add it to the `AppModule` and that's it!

Left to do:

 - Slack widgets
 - Styling

Gif |
-- |
lame-dashboards |

Paired with @florianmski


ouchadam:
something @charroch mentioned to me, to only use `const` unless we need to modify a variable

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
```diff
@@ -1,51 +1,64 @@
...
+  constructor(socketService: SocketService) {
+    this.socketService = socketService;
   }
 
   ngOnInit(): void {
-    this.animation = Observable
-      .timer(DashboardCarouselComponent.ROTATE_RATE_IN_MILLISECONDS, DashboardCarouselComponent.ROTATE_RATE_IN_MILLISECONDS)
-      .subscribe(() => {
-        this.showNextDashboard();
+    let observable: ConnectableObservable<any> = this.socketService.create('http://localhost:3002');
```
ouchadam: this is the server address? would be useful to be constant at the top of the file maybe?

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
```diff
@@ -1,51 +1,64 @@
...
+    this.connection = observable
+      .subscribe((event: WidgetEvent) => {
```
ouchadam: could you pull this out into its own function pretty pleaseeeee

`.subscribe(onWidgetEvent)`

dashboard-all-the-things/src/app/dashboards/external-url/safe.pipe.ts
```diff
@@ -0,0 +1,13 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { DomSanitizer } from '@angular/platform-browser';
+
+@Pipe({ name: 'safe' })
+export class SafePipe implements PipeTransform {
+
+  constructor(private sanitizer: DomSanitizer) {}
+
+  transform(url) {
+    return this.sanitizer.bypassSecurityTrustResourceUrl(url);
```
ouchadam: lol, should convince @blundell to get us a certifacte

dashboard-all-the-things/src/app/config.service.ts
```diff
@@ -3,7 +3,6 @@ import { Http, Response } from '@angular/http';
 import { ReplaySubject, Observable } from 'rxjs';
 import { Config } from './Config';
 import { ErrorObservable } from 'rxjs/observable/ErrorObservable';
-import { SonarDetails } from './dashboards/sonar-coverage/SonarDetails';
 
 @Injectable()
 export class ConfigService {
```
ouchadam: is the class still needed?

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
```diff
@@ -1,51 +1,64 @@
...
+  { key: "stackoverflow", type: StackOverflowComponent }
+];
 
 @Component({
   selector: 'app-dashboard-carousel',
   templateUrl: 'dashboard-carousel.component.html',
-  styleUrls: ['dashboard-carousel.component.scss']
+  styleUrls: ['dashboard-carousel.component.scss'],
+  entryComponents: COMPONENTS.map((el) => el.type)
+    .filter((elem, index, arr) => arr.indexOf(elem) === index)
```
frapontillo: Doesn't this always return `true`?

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
```diff
@@ -1,51 +1,64 @@
...
+import { ReviewComponent } from '../dashboards/review';
+import { SonarCoverageComponent } from '../dashboards/sonar-coverage';
+import { StackOverflowComponent } from '../dashboards/stackoverflow/stackoverflow.component';
+import { WidgetEvent } from '../dashboards/WidgetEvent';
+import { DynamicComponent } from './dynamic.component';
+import { SocketService } from './socket.service';
+import { Component, OnDestroy, OnInit, ViewChild } from '@angular/core';
+import { ConnectableObservable, Subscription } from 'rxjs';
+
+let COMPONENTS = [
```
frapontillo: `const`, `const` everywhere!

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
frapontillo: `const`

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
```diff
@@ -1,51 +1,64 @@
...
-    } else {
-      this.position = 1;
-    }
-    this.setContributorPosition(this.position);
-  }
-
-  private setContributorPosition(position: number) {
-    this.element.nativeElement.style.setProperty(this.CAROUSEL_CURRENT_POSITION, position);
+  findComponentFor(key: string) {
+    return COMPONENTS.filter((elem, index, arr) => elem.key === key)[0];
```
frapontillo: Instead of doing this, why not having the `COMPONENTS` as a dictionary?

```typescript
const COMPONENTS = {
  coverage: SonarCoverageComponent,
  ciWall: ExternalUrlComponent,
  reviews: ReviewComponent,
  stackoverflow: StackOverflowComponent
};
```

dashboard-all-the-things/src/app/dashboard-carousel/dynamic.component.ts
```diff
@@ -0,0 +1,61 @@
...
+
+@Component({
+  selector: 'app-dynamic',
+  template: `<div #target></div>`
+})
+export class DynamicComponent implements OnChanges, AfterViewInit, OnDestroy {
+  @ViewChild('target', { read: ViewContainerRef }) target;
+  @Input() type;
+  @Input() event: WidgetEvent;
+  cmpRef: ComponentRef<DashboardComponent>;
```
frapontillo: Extend the name to be `componentRef` so it's more readable in code.

dashboard-all-the-things/src/app/dashboard-carousel/socket.service.ts
```diff
@@ -0,0 +1,33 @@
...
+  private createObservable(url): ConnectableObservable<any> {
+    let socket = io.connect(url, {
+      'reconnection': true,
+      'reconnectionDelay': 3000,
+      'reconnectionAttempts': 100
+    });
+
+    let observable: Observable<any> = Observable.create((observer: Observer<any>) => {
+      // The Observable will not complete or fail since we want to reconnect if possible.
+      socket.on('message', (event) => {
```
frapontillo: Or `socket.on('message', observer.next);`

dashboard-all-the-things/src/app/dashboards/external-url/external-url.component.html
```diff
@@ -0,0 +1 @@
+<iframe class="frame" [src]="url | safe" *ngIf="url != undefined"></iframe>
```
frapontillo: 😷 but needed for the CI wall I guess?

dashboard-all-the-things/src/app/dashboards/external-url/safe.pipe.ts
frapontillo: 💣 hax it til you make it

dashboard-all-the-things/src/app/dashboards/review/review.component.ts
```diff
@@ -0,0 +1,20 @@
...
+  selector: 'app-review',
+  templateUrl: 'review.component.html',
+  styleUrls: ['review.component.scss']
+})
+export class ReviewComponent implements DashboardComponent {
+
+  private review: Review;
+
+  public update(event: WidgetEvent) {
+    let selection = event.payload.motivators[Math.floor((Math.random() * event.payload.motivators.length))];
```
frapontillo: Extract the random index to a function maybe?

dashboard-all-the-things/src/app/dashboards/stackoverflow/stackoverflow.component.ts
```diff
@@ -0,0 +1,26 @@
...
+  private questions: Array<Question>;
+  private questionToShow: Question;
+
+  public update(event: WidgetEvent) {
+    this.questions = event.payload.questions;
+    this.questionToShow = this.pickRandomQuestion();
+  }
+
+  private pickRandomQuestion(): Question {
+    let question = this.questions[Math.floor(Math.random() * (this.questions.length))];
```
frapontillo: Since this logic is duplicated maybe you want to create a Service that extracts a random Object from an Array?

dashboard-all-the-things/src/app/dashboard-carousel/dynamic.component.ts
```diff
@@ -0,0 +1,61 @@
...
+} from '@angular/core';
+import { DashboardComponent } from './../dashboards/DashboardComponent';
+import { WidgetEvent } from './../dashboards/WidgetEvent';
+import { OnDestroy } from '@angular/core';
+
+@Component({
+  selector: 'app-dynamic',
+  template: `<div #target></div>`
+})
+export class DynamicComponent implements OnChanges, AfterViewInit, OnDestroy {
```
frapontillo: Looks like it is. This allows to set one fixed element in HTML and then change the type at runtime, which will swap view and controller.

dashboard-all-the-things/src/app/dashboards/external-url/safe.pipe.ts
frapontillo: This shouldn't be a certificate issue, more of an Angular block over injecting DOM.

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
frapontillo: This should actually go in the `config.json` so it can be changed when deployed (withouth messing with minified files).

xrigau:
was @florianmski 's work too 👏 

we da web hax0rs

dashboard-all-the-things/src/app/dashboards/external-url/external-url.component.html
ouchadam: for now, we can hit the api instead and renderer it ourselves in the future!

dashboard-all-the-things/src/app/dashboards/external-url/safe.pipe.ts
florianmski: @frapontillo yes that was it...

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
florianmski: For now yes.
In the future the idea is to have multiple widgets rendered by the same component so we need to make sure to have a list with only unique component.
Idea would be that the server would sent a payload that might contains a mix of texts and images and will ask for a template to render those infos, the client shouldn't be aware if the message is about sonar or slack but just on how the infos should be rendered.
Anyway this might be YAGNI then for now :)

dashboard-all-the-things/src/app/dashboards/external-url/external-url.component.html
florianmski: Yep that was also our thinking

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.html
```diff
@@ -51,7 +51,5 @@
   </div>-->
 
   <div id="element" class="carousel">
-    <p class="component">first element</p>
-    <p class="component">second element</p>
-    <p class="component">third element</p>
+    <app-dynamic class="component" [type]="type" [event]="event"></app-dynamic>
```
eduardb: Can't you collapse the empty tag here?

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
```diff
@@ -1,51 +1,64 @@
...
-      .subscribe(() => {
-        this.showNextDashboard();
+    let observable: ConnectableObservable<any> = this.socketService.create('http://localhost:3002');
+    this.connection = observable
+      .subscribe((event: WidgetEvent) => {
+        if (event.widgetKey === undefined) {
+          return;
+        }
+        let res = this.findComponentFor(event.widgetKey);
+        if (res != undefined) {
```
eduardb: What about 
``` 
if (res === undefined) {
    return;
}
```
?
To be consistent with the above `if`

dashboard-all-the-things/src/app/dashboards/external-url/external-url.component.html
eduardb: Collapse to empty tag here as well?

dashboard-all-the-things/src/app/dashboard-carousel/dynamic.component.ts
frapontillo: Hiding and showing would have been crazy and I would have ❌'d this 😆

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
frapontillo: Well, you can get an Object/dictionary/map keys with `Object.keys(COMPONENTS)` and values with `Object.values(COMPONENTS)`.

MAGIC. 🎩

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
takecare: no better names than `el` and `elem`?

dashboard-all-the-things/src/app/dashboard-carousel/dynamic.component.ts
```diff
@@ -0,0 +1,61 @@
...
+    this.updateComponent();
+  }
+
+  ngAfterViewInit() {
+    this.isViewInitialized = true;
+    this.updateComponent();
+  }
+
+  ngOnDestroy() {
+    if (this.cmpRef) {
```
takecare: no `this.cmpRef != undefined`?

frapontillo: What about `null` then? xD

takecare: you'zright!

dashboard-all-the-things/src/app/dashboard-carousel/dynamic.component.ts
florianmski: Yeah there was some started work on this, I'm not sure who started it but we removed it and went with the client being as dumb as possible 👍

dashboard-all-the-things/src/app/config.service.ts
xrigau: yeah this is used to retrieve the github-reports API and will also be used for the server address (https://github.com/novoda/spikes/pull/179#discussion_r86375716)

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
xrigau: @takecare the scope of that function is so small that having a longer name wouldn't really add any value, but a bit of noise, we'll use the same for both  tho 👍

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
xrigau: >[ts] Property 'values' does not exist on type 'ObjectConstructor'.

😢

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.html
xrigau: >Uncaught Error: Template parse errors:

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
xrigau: done!

xrigau:
All comments have been addressed/replied, can we haz re-review please 🙏

dashboard-all-the-things/src/app/dashboard-carousel/dashboard-carousel.component.ts
frapontillo: 1) It's not Node, 😆 
2) Object.values is a valid JS function, hack it big time with `Object['values'](COMPONENTS)` because it's Friday and we can't be bothered

frapontillo: Also, #yolo

xrigau: we still need the values to be unique, how would you go about that?

xrigau: I guess we can still use the filter 🎉

xrigau: done! 8187c73

PR #180, opened 2016-11-03
Enews update ❤️

ouchadam:
Enews wasn't being correctly parsed due to the room name having changed.

Also took the time to rewrite all of the usages so that they're more up to date with my latest javascript knowledge 😉

### DIFF
--- a/enews/enews-node/demo.js
+++ b/enews/enews-node/demo.js
@@ -3,6 +3,6 @@ const token = process.env.token;
 
 var enewsFetcher = new EnewsFetcher(token);
 
-enewsFetcher.getLastSevenDays(function(eNews) {
+enewsFetcher.getLastSevenDays().then(function(eNews) {
   console.log(eNews);
 });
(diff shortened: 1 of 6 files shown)


enews/enews-node/demo.js
```diff
@@ -3,6 +3,6 @@ const token = process.env.token;
 
 var enewsFetcher = new EnewsFetcher(token);
 
-enewsFetcher.getLastSevenDays(function(eNews) {
+enewsFetcher.getLastSevenDays().then(function(eNews) {
```
blundell: i like it 👍

enews/enews-node/enews-fetcher.js
```diff
@@ -1,91 +1,71 @@
...
-    var promises = [];
-    userIds.forEach(function(userId) {
-      promises.push(getUser(userId));
-    });
-    return promises;
-  }
+function filterNonNews(messages) {
+  return Promise.resolve(messages.filter(message => {
+    const text = message.text;
+    return text.includes('#enews') || text.includes('#eNews') || text.includes('#C1V389HGB|enews') || text.includes('#C0YNBKANM');
```
ouchadam: these are what we consider `enews` messages

PR #183, opened 2016-11-05
Complete diff dependency task Gradle plugin

frapontillo:
## Problem

In a Novoda project, we use a custom-made groovy script to analyse the changed files in a PR, so that Jenkins can save time and resources by running:

* Android tests only, if something in the `android` directory has changed
* Apple tests only, if something in the `apple` directory has changed
* All tests if something else has changed

As Novoda is moving more and more towards multi-platform projects, we need a way to favour the use of one Github repository for both Apple and Android projects. At the same time, build times on the CI have to be as short as possible.

## Solution

Most of the times, Pull Requests are tailored to one platform only. In these cases we need to avoid wasting CI build times to test platform-specific modules that haven't changed.

This Gradle plugin uses the command line `git` and the `ghprbTargetBranch` environment variable to check what files have been changed in the current PR (thus relying on the Github PRB Jenkins plugin). The behaviour is still swappable with another one if need be (see `README` and `sample` project).

Optional task dependencies can be defined with an increased Gradle `Task` API:

```gradle
task('testAll')
    .dependsOn('prepareTest') // regular dependency                       
    .onDiffDependsOn(~'android/.*', 'testAndroid')
    .onDiffDependsOn(~'apple/.*', 'testApple')
    .onDiffDependsOn([~'apple/fastlane', ~'android/build.gradle'], 'sync', 'emailDevOps')
```

## Test

Not (yet) tested, but decomposed so that it is easier to test and replace everything.


gradle-diff-task-dependencies/README.md
```diff
@@ -0,0 +1,47 @@
+Gradle Diff Dependency Task
```
frapontillo: The project name, as much as the package and gradle plugin name, suck.
I am open to rename the plugin and main packages/classes to something clearer.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependency.groovy
```diff
@@ -0,0 +1,33 @@
...
+
+import org.gradle.api.Project
+import org.gradle.api.Task
+
+import java.util.regex.Pattern
+
+public class ConditionalDependency {
+
+    static {
+        Project.metaClass.registerConditionalDependency << { ConditionalDependency conditionalDependency ->
```
frapontillo: Only this class uses this custom method, so adding it here will make it possible to test this class without running custom registrations.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependencyEvaluator.groovy
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        matchedDependencies.forEach {
+            it.task.dependsOn(it.dependentTasks)
+        }
+    }
+
+    private static boolean changedFilesContainPattern(List<String> changedFiles, Pattern pattern) {
+        changedFiles.findResult { file ->
+            file.matches(pattern) ? true : null
```
frapontillo: `findResult` stops looping as soon as it returns "anything", and `false` is a valid return value, while `null` isn't.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskConfig.groovy
```diff
@@ -0,0 +1,11 @@
+package com.novoda.gradle.diffdependencytask
+
+import com.novoda.gradle.diffdependencytask.git.ChangedFilesProvider
+
+class DiffDependencyTaskConfig {
+
+    public GroovyCallable<List<String>> changedFilesProvider = ChangedFilesProvider.create()
+
+    public ConditionalDependencyRepository conditionalDependencyRepository = new MemoryConditionalDependencyRepository()
```
frapontillo: The repository is swappable too, for testing reasons.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskPlugin.groovy
```diff
@@ -0,0 +1,28 @@
...
+import com.novoda.gradle.diffdependencytask.ext.ExtensionRegister
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+
+class DiffDependencyTaskPlugin implements Plugin<Project> {
+
+    @Override
+    void apply(Project project) {
+
+        ExtensionRegister.INSTANCE.run()
```
frapontillo: This is needed to make the `Task::ifDiffMatches` method available as soon as the plugin is applied.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ext/ExtensionRegister.groovy
```diff
@@ -0,0 +1,27 @@
...
+import com.novoda.gradle.diffdependencytask.ConditionalDependency
+import org.gradle.api.Task
+
+import java.util.regex.Pattern
+
+enum ExtensionRegister implements Runnable {
+
+    INSTANCE
+
+    private ExtensionRegister() {
```
takecare: is there really a need to have this?

frapontillo: Sadly, yes. Apparently Groovy extension methods don't work on Gradle files through plugins.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependency.groovy
mr-archano: I'm confused. If you want to add a method why not using the extension properties directly (ie: `project.ext`)? `Project` and `Task` are subtypes of [`ExtensionAware`](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.ExtensionAware.html)

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependency.groovy
```diff
@@ -0,0 +1,33 @@
...
+            def config = delegate.extensions.getByName(Constants.TASK_CONFIG_PROPERTY_NAME) as DiffDependencyTaskConfig
+            config.conditionalDependencyRepository.add(conditionalDependency)
```
mr-archano: I believe this is not the most idiomatic way to store your dependencies in Gradle. An option would be to use a [`NamedDomainObjectContainer`](https://docs.gradle.org/current/javadoc/org/gradle/api/NamedDomainObjectContainer.html) in your `project`.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependencyEvaluator.groovy
```diff
@@ -0,0 +1,39 @@
+package com.novoda.gradle.diffdependencytask
+
+import java.util.regex.Pattern
+
+class ConditionalDependencyEvaluator implements GroovyCallable<Void> {
```
mr-archano: out of curiosity: why `GroovyCallable<Void>`? what does it buy you?

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/Constants.groovy
```diff
@@ -0,0 +1,6 @@
+package com.novoda.gradle.diffdependencytask
+
+class Constants {
+    public static String TASK_CONFIG_PROPERTY_NAME = 'diffDependency'
```
mr-archano: If you would manage to remove the `static` block in `ConditionalDependency` you could get rid of this altogether

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskConfig.groovy
mr-archano: but I see no tests tho :trollface:

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ext/ExtensionRegister.groovy
```diff
@@ -0,0 +1,27 @@
...
+        // private ctor
+    }
+
+    @Override
+    void run() {
+        registerTaskExtension()
+    }
+
+    private static void registerTaskExtension() {
+        Task.metaClass.ifDiffMatches << { Pattern[] patterns ->
```
mr-archano: As mentioned above `Task` as subtype of [`ExtensionAware`](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.ExtensionAware.html) owns an `ExtensionProperties`, therefore you should be able to approach this differently. You could for instance amend all the tasks (including the future ones added) via
```
project.tasks.all { task -> /* blah*/ }
```

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/git/CurrentBranchProvider.groovy
```diff
@@ -0,0 +1,11 @@
+package com.novoda.gradle.diffdependencytask.git
+
+class CurrentBranchProvider implements GroovyCallable<String> {
```
mr-archano: same question here: why `GroovyCallable`?

gradle-diff-task-dependencies/sample/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+task('testApple') << {
+    println 'This is an apple test task.'
+}
+
+task('selectiveTest') << {
+    println 'This will run after 0 or more between "testAndroid" and "testApple".'
+}
+
+tasks.getByName('selectiveTest')
+        .ifDiffMatches(~'android/.*').dependsOn('testAndroid')
```
mr-archano: I have to say I'm not big fan of this API: why do you need to break the chain in 2 steps (`ifDiffMatches()` + `dependsOn`)? Is `ifDiffMatches()` considered a standalone operation? To me it seems like you always have to complete via `dependsOn()` (that by the way might be confusing if you're used to `Task#dependsOn`), so why not having a single call with both parameters? You could use varargs for tasks if you want to add multiple conditional depedencies

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependencyEvaluator.groovy
frapontillo: It's simply my noobiness at Groovy 😆. Is there an `Action` equivalent in Groovy? The interface is there so that it can be tested with any implementation.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskConfig.groovy
frapontillo: Yet! 😛

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/git/CurrentBranchProvider.groovy
frapontillo: Same answer as before!

gradle-diff-task-dependencies/sample/build.gradle
frapontillo: I'd rather have a well defined API over:
* named containers on `ext` (so that the dependency declaration is similar to the regular one)
* `dependsOn` with patterns and task names (as the original `dependsOn` already uses varargs of `Object`, discriminating between the two could be messy if possible at all)

These were really the only two reasons that pushed me to go for the "code over configuration" approach.

frapontillo:
Good points raised by @mr-archano, I will address them later. 💯

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependencyEvaluator.groovy
frapontillo: Yes, of course!

gradle-diff-task-dependencies/sample/build.gradle
mr-archano: @frapontillo I'm not sure I fully understand your last comment. I feel like a chat could be better than GitHub in this case 👍

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependency.groovy
frapontillo: But wouldn't this expose the container to project configuration? That's something I would like to avoid.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/Constants.groovy
frapontillo: Done in de93a16128516dce496f6ffe9d867e4c8d7b68cc.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/ext/ExtensionRegister.groovy
frapontillo: Done in e289d4d89eed9670882e70e6bcd99d2dcb39149c.

frapontillo:
@mr-archano main comments have been addressed, could we have a chat in the afternoon to address the API issue?

frapontillo:
Code changed to reflect @mr-archano's ~~biased~~ experienced Groovy/Gradle knowledge.

😛

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskPlugin.groovy
```diff
@@ -0,0 +1,46 @@
...
+import java.util.regex.Pattern
+
+class DiffDependencyTaskPlugin implements Plugin<Project> {
+
+    private static String TASK_CONFIG_PROPERTY_NAME = 'diffDependency'
+
+    @Override
+    void apply(Project project) {
+
+        project.extensions.add(TASK_CONFIG_PROPERTY_NAME, new DiffDependencyTaskConfig())
```
mr-archano: This is not used outside this class now, right? You can just new it as local variable and pass the reference around.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskConfig.groovy
mr-archano: I still fail to see how this helps testability: this object is only holding 2 fields that could be inside the plugin itself 😕
(I second the choice of Liskov-ing them tho)

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskPlugin.groovy
```diff
@@ -0,0 +1,46 @@
...
+
+        project.ext.registerConditionalDependency = { ConditionalDependency conditionalDependency ->
```
mr-archano: This method is now used only inside this class too. You could remove it altogether.

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskPlugin.groovy
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        project.afterEvaluate {
+            def config = project.extensions.getByName(TASK_CONFIG_PROPERTY_NAME) as DiffDependencyTaskConfig
+            def evaluator = new ConditionalDependencyEvaluator(
+                    config.changedFilesProvider,
+                    config.conditionalDependencyRepository
+            )
+
+            evaluator()
```
mr-archano: Is having `()` the only reason why the evaluator class extends from `GroovyCallable`? You could create this object before the `afterEvaluate` block and then have only this line here 💃

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/git/ChangedFilesProvider.groovy
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    ChangedFilesProvider(GroovyCallable<String> currentBranchProvider,
+                         GroovyCallable<String> targetBranchProvider) {
+
+        this.getCurrentBranch = currentBranchProvider
+        this.getTargetBranch = targetBranchProvider
+    }
+
+    private synchronized List<String> getChangedFiles() {
```
mr-archano: I was wondering why the `synchronized` here (and below). Are you afraid this will be executed in concurrent threads?

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskPlugin.groovy
frapontillo: It can be used by clients, see https://github.com/novoda/spikes/pull/183/files#diff-fb0c667cda520bf910e0ca1878413fd9R39

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskPlugin.groovy
frapontillo: If I create it before `afterEvaluate` I cannot access the user-custom configuration right?

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskConfig.groovy
frapontillo: We can inject whatever changed files provider and repository in here and test the plugin with mock data and behaviour. Isn't this the only way to configure a plugin?

frapontillo:
@takecare do you prefer the new API or the old one?

takecare:
@frapontillo you mean `onDiffDependsOn` vs `ifDiffMatches`?

frapontillo:
Yes, meaning we now have a single method instead of a builder-like chain of calls.

takecare:
i feel the older/original approach was more readable but this one ain't bad either. no strong feelings atm 😄

gradle-diff-task-dependencies/plugin/src/main/groovy/com/novoda/gradle/diffdependencytask/DiffDependencyTaskConfig.groovy
mr-archano: You're right. The convention is to use the `*Extension` suffix in this case tho. That and reviewing gradle code too late was the cause of my comment. You can just ignore it ;)

takecare:
>  9 days ago

time to merge!

PR #185, opened 2016-11-08
Slacker backend tidy up

ouchadam:
- Moves all of the client code to a client directory
- Starts fixing the code styles to match javascript conventions (eg prototype pattern)
- Uses a `config.json` instead of relying on enviroment variables 

Config example
```json
{
  "server": {
    "port": ""
  },
  "widgets": {
    "sonar": {
      "token": ""
    },
    "slack": {
      "token": ""
    }
  }
}
```


xrigau:
💯

PR #186, opened 2016-11-08
PR-272/PT-298: Use testing verb correctly in custom subjects for static-analysis-plugin

mr-archano:
> Tracked in JIRA by [PT-298](https://novoda.atlassian.net/browse/PT-298)

### Scope of the PR

We are using [`Truth`](https://github.com/google/truth) as assertion framework in our tests. This allows us to offering extensions to the assertions API tailored to the targets of our tests. We noticed in few places the implementation was incorrectly forcing assertions defeating the whole ["_We've made failure a strategy_"](https://github.com/google/truth#truth) mantra.

### Considerations/Implementation Details

- Amended the custom `Subject`s to retrieve the current testing verb (ie: `assert` vs `assume`) instead of enforcing one in the implementation. This is possible via [`Subject#check()`](http://google.github.io/truth/api/latest/com/google/common/truth/Subject.html#check()), that allows to chain other `Subject` within our custom propositional logic
- Removed custom failure in `TestProjectSubject` in favour of the simple [`Subject#instanceOf()`](http://google.github.io/truth/api/latest/com/google/common/truth/Subject.html#isInstanceOf(java.lang.Class)) check available out-of-the-box.

#### Testing

Refactoring only. We checked tests are still passing.

> Paired with @eduardb

### DIFF
--- a/static-analysis-plugin/plugin/src/test/groovy/com/novoda/test/LogsSubject.groovy
+++ b/static-analysis-plugin/plugin/src/test/groovy/com/novoda/test/LogsSubject.groovy
@@ -1,9 +1,11 @@
 package com.novoda.test
 
 import com.google.common.truth.FailureStrategy
+import com.google.common.truth.StringSubject
 import com.google.common.truth.Subject
 import com.google.common.truth.SubjectFactory
 import com.google.common.truth.Truth
+import com.google.common.truth.TruthJUnit
 
 import javax.annotation.Nullable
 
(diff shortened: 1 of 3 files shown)


static-analysis-plugin/plugin/src/test/groovy/com/novoda/test/TestProjectSubject.groovy
```diff
@@ -25,10 +25,6 @@ class TestProjectSubject extends Subject<TestProjectSubject, TestProject> {
     }
 
     public void isAndroidProject() {
-        if (!(actual() instanceof TestAndroidProject)) {
-            failureStrategy.fail(String.format('project was instance of <%s>, but was instance of <%s> instead',
-                    TestAndroidProject.simpleName,
-                    actual().getClass().simpleName))
-        }
+        check().that(actual()).isInstanceOf(TestAndroidProject)
```
takecare: 😎

static-analysis-plugin/plugin/src/test/groovy/com/novoda/test/LogsSubject.groovy
```diff
@@ -25,51 +27,56 @@ class LogsSubject extends Subject<LogsSubject, Logs> {
         Truth.assertAbout(FACTORY).that(logs)
     }
 
+    public static LogsSubject assumeThat(Logs logs) {
```
mr-archano: I suppose this is a side-effect of being a Java programmer, where defaults are different. I don't mind to push for something more idiomatic, but I wonder if in this case it wouldn't make more sense to be explicit to identify public APIs.

static-analysis-plugin/plugin/src/test/groovy/com/novoda/test/PenaltyExtensionSubject.groovy
```diff
@@ -23,20 +22,25 @@ class PenaltyExtensionSubject extends Subject<PenaltyExtensionSubject, PenaltyEx
...
-    private PenaltyExtensionSubject(FailureStrategy failureStrategy, @Nullable PenaltyExtension actual) {
-        super(failureStrategy, actual)
-    }
-
     public static PenaltyExtensionSubject assertThat(PenaltyExtension extension) {
         assertAbout(FACTORY).that(extension)
     }
 
+
+    public static PenaltyExtensionSubject assumeThat(PenaltyExtension extension) {
```
xrigau: I'm a bit confused... these new public methods aren't being used, is that gonna be part of the next PR?

xrigau:
Just 1 Question but not blocking so :shipit:

static-analysis-plugin/plugin/src/test/groovy/com/novoda/test/LogsSubject.groovy
takecare: well, i asked because everything being public, the usage of private would work as documentation. anyway, not a biggie.

static-analysis-plugin/plugin/src/test/groovy/com/novoda/test/PenaltyExtensionSubject.groovy
mr-archano: No they aren't used, it was just to showcase how this can evolve. I was showing this to @eduardb and then we decided to keep it (despite YAGNI & co).

PR #187, opened 2016-11-08
Slacker: Fix using token instead of apiKey for sonar

florianmski:
#185 introduced a `config.json` file where we can dump tokens instead of exporting the variables.

There was a slight error where we where getting back from the `config.json` file an `apiKey` instead of `token` for the sonar part.

Fixed it ✨


ouchadam:
oops my bad 👍

xrigau:
yeah nice one @florianmski 🎉

PR #189, opened 2016-11-08
Include TechDebt annotation + processor 

juankysoriano:
#### Problem

This PR introduces a `@TechDebt` annotation, alongside with a `TechDebtAnnotationProcessor.java` which will output the following as a warning whenever a `@TechDebt` annotation is found. The format is like this:

```
/path/YourClass.java:10: warning: TechDebt detected!!!
public class YourClass {
       ^
  Description: We need to solve this because bla and ble.
  Size: LARGE
  Link: http://link-to-github/
```

The end goal is to combine this with the CI Warnings plugin in order to achieve something like this:

image

image


TechDebtsTracking/core/src/main/java/com/novoda/support/annotations/TechDebt.java
```diff
@@ -0,0 +1,32 @@
...
+
+    enum Size {
+        XSMALL,
+        SMALL,
+        MEDIUM,
+        LARGE,
+        XLARGE
+    }
+
+    Size size();
```
eduardb: Shouldn't have been `Size` instead of `String`?

TechDebtsTracking/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Sat Feb 06 01:19:58 GMT 2016
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.5-all.zip
```
eduardb: Why not a newer version, like `2.14.1` or `3.1`?

TechDebtsTracking/core/src/main/java/com/novoda/support/annotations/TechDebt.java
```diff
@@ -0,0 +1,32 @@
...
+        ElementType.FIELD,
+        ElementType.LOCAL_VARIABLE,
+        ElementType.METHOD,
+        ElementType.PACKAGE,
+        ElementType.PARAMETER,
+        ElementType.TYPE})
+public @interface TechDebt {
+
+    enum Size {
+        XSMALL,
```
mr-archano: out of curiosity - what's the rationale of introducing this dimension? how would you decide the size of tech debt?

TechDebtsTracking/core/src/main/java/com/novoda/support/annotations/TechDebt.java
mr-archano: Isn't `Size` already? 😅

eduardb: Wth. I am pretty sure that it was `String` previously 😂 My bad.

juankysoriano: Uhmmm 😄

TechDebtsTracking/gradle/wrapper/gradle-wrapper.properties
juankysoriano: https://github.com/novoda/spikes/pull/189/commits/fa62165573fad033038562438d0fe54b246fa6cf

mr-archano:
Other question: could you detail how this integrates into Jenkins? We are discussing how to make the whole build setup self-contained in our build scripts to reduce the maintenance of our CI jobs and enabling scaling etc. I also would be interested to know why the choice of an annotation processor over some facility to inspect source files instead (_à la_ Checkstyle, PMD, etc).

juankysoriano:
@mr-archano I will answer your questions with more details, but what I can tell you now is that the `README.md` here is gonna be improved with a way more detailed information about:

1. The intention.
2. Use cases.
3. How to integrate this with Jenkins.

Regarding to:

> It would be cool to have a sample project and possibly some integration test for this. I would consider a simple gradle plugin to actually apply the pre-processor, to provide a simple, setup-less, drop-in solution.

Sample project will be provided, will think about the value of integration tests and about the `gradle` plugin, if you offer your help I am 100% in for it.

TechDebtsTracking/core/src/main/java/com/novoda/support/annotations/TechDebt.java
juankysoriano: The idea of the annotation is to encourage two things: 

- Tech debts monitoring
- Tech debts addressing. 

Given the second, `Tech debts addressing` `XSMALL` would represent a tech-debt which can be very easily addressed, (5-15 minutes?)so a developer can boyscout it as part of his work if he ever find one of those while working on a certain area.

This was the original idea, of course this PR wouldn't make sense if I am not open to suggestions, so please shout if you have a better idea :D

juankysoriano:
Closing while I work on a sample project and Readme enhancements.

juankysoriano:
@mr-archano @eduardb I have elaborated a `Readme.md` plus included a small sample.

https://github.com/novoda/spikes/blob/8593e282f6722f10861affb206e9f35f64c090e2/TechDebtsTracking/README.md

I think we can continue the discussion. 

To be honest, @mr-archano when it gets about how to automatize the integration with jenkins, or the possibility of including this on a gradle plugin I get a bit lost...

juankysoriano:
Yeah @eduardb there should be possible to create a custom version of the jenkins warnings plugins to include the tech-debt parser already embedded, however there will be still a manual step of installing the plugin on your jenkins instance.

TechDebtsTracking/core/src/main/java/com/novoda/support/annotations/TechDebt.java
danybony: I like the idea of having an indication about the size.
I would keep it simple though, with just SMALL, MEDIUM and LARGE for a rough estimation.
It's true that having more options is just giving more flexibility, so no problem 👍

juankysoriano:
Thanks for the review @danybony

juankysoriano:
@eduardb @danybony @mr-archano We have now a fully working sample project, you can see the output in the CI in the following link:

https://ci.novoda.com/job/tech-debt-tracking/20/

eduardb:
It's time to 🚢

PR #190, opened 2016-11-08
Introducing the BotRemote activity! 🤖

alexstyl:
This PR introduces the BotControllerActivity which will be used to show a controller to move the bot around. It includes a separate app icon and an activity icon as we might have multiple launchers within the same app.


TelepresenceBot/mobile/src/main/AndroidManifest.xml
```diff
@@ -1,14 +1,19 @@
...
+    android:icon="@mipmap/ic_launcher"
     android:label="@string/app_name"
     android:supportsRtl="true"
-    android:theme="@style/AppTheme">
-    <activity android:name=".TpBotActivity">
+    android:theme="@style/AppTheme"
+    tools:ignore="AllowBackup,GoogleAppIndexingWarning">
+    <activity
+      android:name=".BotControllerActivity"
+      android:icon="@mipmap/ic_bot_controller_launcher"
```
danybony: 👍

PR #193, opened 2016-11-09
PT-294: Integrate bintray-release plugin to publish static-analysis-plugin artifact

mr-archano:
> Tracked in JIRA by [PT-294](https://novoda.atlassian.net/browse/PT-294)

### Scope of the PR

We want to integrate the plugin in some of the running projects in Novoda. For this we need to publish an artifact to Bintray.

### Considerations/Implementation Details

The publishing configuration will use the [`bintray-release`](https://github.com/novoda/bintray-release/) plugin. The current version of the plugin is set to `0.1`. The artifact will be deployed via our Jenkins as usual.

#### Testing
Manually checked the artifact is created correctly along with the `pom.xml` when the `bintrayUpload` task is executed.

> Paired with @eduardb

### DIFF
--- a/static-analysis-plugin/build.gradle
+++ b/static-analysis-plugin/build.gradle
@@ -3,6 +3,9 @@ subprojects {
         repositories {
             jcenter()
         }
+        dependencies {
+            classpath 'com.novoda:bintray-release:0.4.0'
+        }
     }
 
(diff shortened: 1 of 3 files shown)


xrigau:
Good stuff! :shipit:

PR #213, opened 2016-12-01
TpBot: Add a TextViewWithForeground

Mecharyry:
#### Problem
We have a landing screen but the clickable items do not have any touch states. 

#### Solution
Rather than wrapping the items in a view that handles foreground (FrameLayout) we'll create a `TextView` that can support a `foreground`.

#### Screen Capture
tp_foreground mp4

### DIFF
--- a/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/TextViewWithForeground.java
+++ b/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/TextViewWithForeground.java
@@ -39,16 +39,19 @@ public class TextViewWithForeground extends TextView {
             if (resourceId == INVALID_RESOURCE_ID) {
                 return;
             }
-            final Drawable drawable = getDrawable(resourceId);
+            final Drawable drawable = getDrawableFor(resourceId);
             updateForegroundWith(drawable);
-            createHotspotTouchListener();
+
+            if (isLollipopOrAfter()) {
+                createHotspotTouchListener();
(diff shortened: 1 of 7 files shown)


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/TextViewWithForeground.java
```diff
@@ -0,0 +1,136 @@
...
+    private static Drawable getDrawableLollipop(Resources resources, @DrawableRes int resId) {
+        return resources.getDrawable(resId, null);
+    }
+
+    private static Drawable getDrawablePreLollipop(Resources resources, @DrawableRes int resId) {
+        return resources.getDrawable(resId);
+    }
+
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    private void createHotspotTouchListener() {
```
danybony: You are always calling this, even with pre-lollipop devices

TelepresenceBot/mobile/src/main/res/values/styles.xml
```diff
@@ -14,11 +14,13 @@
 
   <style name="Landing.Human">
     <item name="android:background">@android:color/holo_blue_light</item>
+    <item name="foreground">@drawable/landing_option_foreground</item>
```
danybony: This could go in the common `Landing` parent theme

danybony:
Good one 👍

PR #214, opened 2016-12-01
TpBot: Result

Mecharyry:
#### Problem
Need to represent a result when attempting to connect to the server (bot) part of the app and vice-versa. 

#### Solution
Create a `Result` that uses `Optional` to represent either a success message or an error exception. This will be implemented in a follow-up PR when the `TpBotService` is added to the project.

### DIFF
--- a/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/Result.java
+++ b/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/Result.java
@@ -1,0 +1,35 @@
+package com.novoda.tpbot;
+
+import com.novoda.tpbot.support.Optional;
+
+public class Result {
+
+    private final Optional<String> message;
+    private final Optional<Exception> exception;
+
+    public static Result from(String message) {
+        return new Result(Optional.of(message), Optional.<Exception>absent());
+    }
+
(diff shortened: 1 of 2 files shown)


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/Result.java
```diff
@@ -0,0 +1,35 @@
...
+    public Optional<String> message() {
+        return message;
+    }
+
+    public Optional<Exception> exception() {
+        return exception;
+    }
+
+    public boolean isError() {
+        return exception.isPresent();
```
danybony: I'm not entirely sure about this solution...
Are we saying that everything without an exception is a success? Even a result without both exception and message?

frapontillo: I guess the whole point would be that when constructing the result, the error logic has already been done. If there is invalid data, there should be an exception.

danybony: Ok, let's keep this one going, we can see in future PRs if we'll have some special cases

PR #216, opened 2016-12-07
TpBot: Observer

Mecharyry:
#### Problem
We want to be able to perform an asynchronous call but keep it as far away from the `Presenter` as possible. 

#### Solution
Implement `Observer` pattern to supply results to the `Presenter` when a new value is emitted by the `TpService`.

The `Presenter` and `Service` are coming up in the next PR so bare with me 👅 This also gave me the opportunity to implement a simple `Observer` rather than using Rx, which I haven't done before 😄

### DIFF
--- a/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
+++ b/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
@@ -17,9 +17,12 @@ public abstract class Observable<T> {
         if (observer == null) {
             throw new DeveloperError("You cannot attach a null observer");
         }
-        if (!observers.contains(observer)) {
-            observers.add(observer);
+
+        if (observers.contains(observer)) {
+            return this;
         }
+
+        observers.add(observer);
 
         return this;
     }
--- a/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
+++ b/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
@@ -29,20 +29,9 @@ public abstract class Observable<T> {
         observers.remove(observer);
     }
 
-    protected void notifyOf(T newValue) {
-        Observer<T>[] observersCopy;
-
-        synchronized (this) {
-            if (hasNotChanged()) {
-                return;
-            }
-
-            observersCopy = observers.toArray(new Observer[observers.size()]);
-            clearChanged();
-        }
-
-        for (int i = observersCopy.length - 1; i >= 0; i--) {
-            observersCopy[i].update(newValue);
+    protected synchronized void notifyOf(T newValue) {
+        for (int i = observers.size() - 1; i >= 0; i--) {
+            observers.get(i).update(newValue);
         }
     }
 
--- a/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
+++ b/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
@@ -13,22 +13,22 @@ public abstract class Observable<T> {
         observers = new ArrayList<>();
     }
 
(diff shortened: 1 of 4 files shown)


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
```diff
@@ -0,0 +1,69 @@
...
+    protected Observable() {
+        observers = new ArrayList<>();
+    }
+
+    public synchronized Observable<T> attach(Observer<T> observer) {
+        if (observer == null) {
+            throw new DeveloperError("You cannot attach a null observer");
+        }
+
+        if (observers.contains(observer)) {
```
frapontillo: Better: if it doesn't contain  the observer add it, then `return this` anyway.

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
```diff
@@ -0,0 +1,69 @@
...
+
+    protected void notifyOf(T newValue) {
+        Observer<T>[] observersCopy;
+
+        synchronized (this) {
+            if (hasNotChanged()) {
+                return;
+            }
+
+            observersCopy = observers.toArray(new Observer[observers.size()]);
```
frapontillo: Why the copy?

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
```diff
@@ -0,0 +1,69 @@
...
+
+    protected synchronized void detachObservers() {
+        observers.clear();
+    }
+
+    protected synchronized void setChanged() {
+        hasChanged = true;
+    }
+
+    private synchronized void clearChanged() {
```
frapontillo: No need to synchronise this and `hasNotChanged` since its usage is already `synchronized`.

Mecharyry: 👍  Did this before I decided to make these private only 😄

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
Mecharyry: Safety I guess. I didn't want someone to remove from the observers while I am trying to update. I guess that is guarded against because of the `synchronized` on the `detach` though so this won't happen. I'll remove 😄

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
Mecharyry: After thinking about this some more I think I would rather keep the `synchronized` it shows intent that this can be used on different threads. Removing it might be misleading to someone in the future.

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Observable.java
Mecharyry: Not sure about this now... If I remove the copy, someone could feasibly remove something whilst I am updating right? the update is not inside a `synchronized` block.

Mecharyry: Instead do this? So it blocks until it has finished doing all of the updates.
``` 
    protected synchronized void notifyOf(T newValue) {
        for (int i = observers.size() - 1; i >= 0; i--) {
            observers.get(i).update(newValue);
        }
    }
```

frapontillo: DAT.

Mecharyry:
Thanks for the review @frapontillo, very much appreciated 😄

PR #218, opened 2016-12-09
TpBot: Bot Presenter

Mecharyry:
#### Problem
`TpBot` has a  nice `BotTpService` and a `BotView` interface but not a presentation layer that binds the two together.

#### Solution
Create the `BotPresenter` to handle the calls to the `TpService` and accept pushes from that service to display in the UI. This is currently just the `BotPresenter` a follow-up PR will introduce the `HumanPresenter` and the associated services.


TelepresenceBot/core/src/main/java/com/novoda/tpbot/support/Observable.java
```diff
@@ -33,22 +36,26 @@ protected synchronized void notifyOf(T newValue) {
...
 
-    protected synchronized void detachObservers() {
+    public synchronized void detachObservers() {
         observers.clear();
     }
 
-    protected synchronized void setChanged() {
-        hasChanged = true;
-    }
+    public abstract Observable<T> start();
```
frapontillo: Can we move the abstract method as the first class method so it's easier to understand what needs implementing?

Mecharyry: I agree, being able to see instantly see what you need to implement is a good idea :+1:

TelepresenceBot/core/src/main/java/com/novoda/tpbot/support/Observable.java
```diff
@@ -4,10 +4,13 @@
 
 public abstract class Observable<T> {
 
-    private boolean hasChanged = false;
     private final ArrayList<Observer<T>> observers;
```
Mecharyry: I hadn't considered that :+1: changed to use a Set 😄 638f7de

PR #239, opened 2016-12-16
Tpbot/List connected USB devices

danybony:
In the future, we will need to know the Product ID and Vendor ID of USB connected Arduino, so that we can whitelist it and start our app automatically when connected.
This PR adds a useful menu item that shows in a toast these detail for every connected USB device (if any).
screenshot_20161216-101044

### DIFF
--- a/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/BotActivity.java
+++ b/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/BotActivity.java
@@ -5,6 +5,9 @@ import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
 import android.view.LayoutInflater;
+import android.view.Menu;
+import android.view.MenuInflater;
+import android.view.MenuItem;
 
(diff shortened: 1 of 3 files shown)


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/BotActivity.java
```diff
@@ -29,4 +37,41 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        } else {
+            for (UsbDevice device : devices.values()) {
+                builder.append(
+                        getString(R.string.usb_device_name_vendor_product,
+                                  device.getDeviceName(),
+                                  device.getVendorId(),
+                                  device.getProductId()
+                        )
+                );
+            }
```
frapontillo: No separator between devices?

danybony: Ah good point! I had it before using the StringBuilder and I forgot to add it back 👍

danybony: Here it is 	16d9bbf

PR #240, opened 2016-12-16
Tpbot/Repeat commands

danybony:
When we will connect the controller UI with the movement unit, it will be good to stop moving the bot as soon as the user releases the control button.
To do so, before we were sending a `onDirectionPressed` and `onDirectionReleased` commands, but I think it will be more reliable to continuously send a direction command while a button is pressed and stop sending as soon as it's released.
On the movement unit we adopt a mirrored mechanism, starting a timeout after each direction command, so that if we keep receiving them we continue in the same direction, otherwise we stop moving.
This should give us more reliability in case of loss of connection for example.

device-2016-12-16-132257

### DIFF
--- a/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/human/HumanActivity.java
+++ b/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/human/HumanActivity.java
@@ -1,6 +1,7 @@
 package com.novoda.tpbot.human;
 
 import android.os.Bundle;
+import android.os.Handler;
 import android.support.v7.app.AppCompatActivity;
 
 import com.novoda.notils.caster.Views;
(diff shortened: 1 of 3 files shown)


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/human/HumanActivity.java
```diff
@@ -1,50 +1,84 @@
...
             @Override
             public void onLazersFired() {
-                debugView.showTimed(LAZERS, DURATION_LAZERS);
+                startRepeatingCommand(LAZERS);
             }
 
             @Override
             public void onLazersReleased() {
-                debugView.clearMessage();
+                stopRepeatingCommand(LAZERS);
```
frapontillo: sio-boppo-laser

PR #243, opened 2016-12-22
Tpbot server declaration

Mecharyry:
#### Problem
The current implementation for the server is relying on `SocketIO` and I am not using a static IP for the machine, which means that the IP address changes between instances of the server. 

#### Solution
Add a way to specify the server we want to connect to in the app. A switchable view is used to swap between the server declaration and the controls view when a connection is established. Currently locked on server declaration, a follow-up PR will add the services to allow the switch to happen automatically when a connection is established.

#### Screen Captures
Before | Now with Padding!
--- | ---
server_connection | tpbot_padding


TelepresenceBot/mobile/build.gradle
```diff
@@ -1,12 +1,12 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 24
-    buildToolsVersion "24.0.2"
+    compileSdkVersion 25
+    buildToolsVersion "24.0.3"
```
tasomaniac: why not buildToolsVersion `25.0.2` since we are now on sdk `25`

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/human/HumanActivity.java
```diff
@@ -7,54 +7,62 @@
...
-            }
-        });
+        ControllerView controllerView = Views.findById(switchableView, R.id.bot_controller_direction_view);
+        controllerView.setControllerListener(controllerListener);
+
+        ServerDeclarationView serverDeclarationView = Views.findById(switchableView, R.id.bot_server_declaration_view);
+        serverDeclarationView.setServerDeclarationListener(serverDeclarationListener);
     }
 
+    private final ControllerListener controllerListener = new ControllerListener() {
```
tasomaniac: I'm not used to field declaration after methods. Is it something we do?

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/human/ServerDeclarationView.java
```diff
@@ -0,0 +1,41 @@
...
+        super.onFinishInflate();
+        View.inflate(getContext(), R.layout.merge_server_declaration, this);
+
+        final EditText serverAddressDeclaration = Views.findById(this, R.id.bot_server_declaration);
+        View connectToServer = Views.findById(this, R.id.bot_server_connect);
+
+        connectToServer.setOnClickListener(new OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                String serverAddress = String.valueOf(serverAddressDeclaration.getText());
```
tasomaniac: I would use `serverAddressDeclaration.getText().toString()` instead.

TelepresenceBot/mobile/src/main/res/layout/activity_human.xml
```diff
@@ -1,24 +1,27 @@
...
     android:id="@+id/bot_controller_debug_view"
     style="@style/Controller.DebugText"
     android:layout_width="match_parent"
     android:layout_height="0dp"
     android:layout_weight="50"
     tools:text="This is some text yo" />
 
-  <com.novoda.tpbot.human.ControllerView
-    android:id="@+id/bot_controller_direction_view"
+  <com.novoda.tpbot.support.SwitchableView
```
tasomaniac: Is `SwitchableView` something we use elsewhere.
Instead of creating a custom view, I would just put the layouts as a children of a `ViewSwitcher` and extract that `ViewSwitcher` as an xml file. 
Advantages would be that you just include that instead of duplicating it for portrait and landscape.

TelepresenceBot/mobile/src/main/res/layout/merge_server_declaration.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<merge xmlns:android="http://schemas.android.com/apk/res/android">
+
+  <EditText
+    android:id="@+id/bot_server_declaration"
+    style="@style/Human.Server"
+    android:layout_width="0dp"
+    android:layout_height="wrap_content"
+    android:layout_weight="75"
```
tasomaniac: `nit:` maybe some `layout_margin` around the edittext and the button :)

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/human/HumanActivity.java
Mecharyry: I tend to do this for interface declarations because then they are directly below the usage, making the code flow better.

TelepresenceBot/mobile/build.gradle
Mecharyry: Will do 👍

TelepresenceBot/mobile/src/main/res/layout/activity_human.xml
Mecharyry: Declaring the layouts as children of the `ViewSwitcher` is what I disliked about it and it **can** make the layout file super difficult to read. In this case, it might not be a problem but personally, I prefer it this way 😕  Would anyone else prefer to see this done another way?

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/human/HumanActivity.java
tasomaniac: :+1: it makes sense. I am used to have methods creating the listener instead of having a field. But most of the time we need the field anyways.

TelepresenceBot/mobile/src/main/res/layout/activity_human.xml
tasomaniac: I didn't say that I prefer to see it done another way 😄 It was just a question.
I agree that `ViewSwitcher` is kinda weird to use. It is much straightforward with your custom view.

TelepresenceBot/mobile/src/main/res/layout/merge_server_declaration.xml
Mecharyry: Done and done faae014

Mecharyry:
Merge away! :-D

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/human/ServerDeclarationView.java
```diff
@@ -0,0 +1,41 @@
...
+        View.inflate(getContext(), R.layout.merge_server_declaration, this);
+
+        final EditText serverAddressDeclaration = Views.findById(this, R.id.bot_server_declaration);
+        View connectToServer = Views.findById(this, R.id.bot_server_connect);
+
+        connectToServer.setOnClickListener(new OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                String serverAddress = serverAddressDeclaration.getText().toString();
+                serverDeclarationListener.onConnect(serverAddress);
```
Mecharyry: 👍  Completely agree, would rather throw than keep the NO_OP, will remove in a follow-up PR.

PR #245, opened 2017-01-02
New Spike: MapIdlingResource

tasomaniac:
This is a very simple IdlingResource that finds the SupportMapFragment within your Activity and waits for its initialization to run the tests.

Useful if you start your display after the map is initialized. 

Since Google Maps start asynchronous and since it doesn't use Android View components, Espresso does not wait for its initialization by default. Anything you do after the map is initialized causes problems with Espresso. 

This implementation is very simple and assumes that you have only 1 map in your Activity.


map-idling-resource/library/src/main/java/com/novoda/espresso/MapIdlingResource.java
```diff
@@ -0,0 +1,78 @@
...
+
+        return MainThread.createOnMainThread(mapFragment);
+    }
+
+    @Nullable
+    private static SupportMapFragment findSupportMapFragment(FragmentManager supportFragmentManager) {
+        List<Fragment> fragments = supportFragmentManager.getFragments();
+        for (Fragment fragment : fragments) {
+            if (fragment instanceof SupportMapFragment) {
+                return (SupportMapFragment) fragment;
```
ouchadam: is it likely to have more than one map fragment? 
Do you think it's worth matching based on the tag? (or is this overkill for a demo spike 😄 )

tasomaniac: Yep, this assumes that you wait for only 1 map initialization. 
There are apps that have multiple maps in 1 screen. But in this case, they are mostly for display purposes. 
Either I should make that assumption very clear. Or I can hook into all MapFragment's available in the `FragmentManager`
What do you think?

ouchadam: I like the sound of making it clearer 👍

tobiasheine:
Might be a candidate for our [Cookbook](https://github.com/novoda/android-demos/tree/master/Cookbook)

tasomaniac:
@blundell Can you take a look?

I opened this PR before the idea to create Cookbook. 
I don't know if this should be a super small library, or should it be a recipe? What do you think?

mr-archano:
@tasomaniac I'd vote for a recipe. The code as it is can be put in `android-demos` too. Nice one btw 👍

tasomaniac:
@mr-archano Being in `android-demos` means that it needs a blog post. That's even better. 👍 
Closing this for now and I will come with a blog post about it.

blundell:
👍 💯

PR #247, opened 2017-01-03
Test gradle-diff-task-dependencies

frapontillo:
The differential task dependencies plugin for Gradle lacks some basic testing for the evaluation logic.

This PR introduces some refactoring that make testing easier without actually building sample Gradle projects.

After this PR is merged, we will release a first version of the plugin, so it can be used in our projects.

### DIFF
--- a/gradle-diff-task-dependencies/plugin/src/test/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependencyEvaluatorTest.groovy
+++ b/gradle-diff-task-dependencies/plugin/src/test/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependencyEvaluatorTest.groovy
@@ -1,0 +1,104 @@
+package com.novoda.gradle.diffdependencytask
+
+import org.junit.Before
+import org.junit.Test
+import org.mockito.InjectMocks
+import org.mockito.Mock
+
+import static com.google.common.truth.Truth.assertThat
+import static org.mockito.Mockito.when
+import static org.mockito.MockitoAnnotations.initMocks
+
+class ConditionalDependencyEvaluatorTest {
+
+    @Mock
+    private ConditionalDependencyRepository dependenciesRepository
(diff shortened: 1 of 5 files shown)


gradle-diff-task-dependencies/plugin/src/test/groovy/com/novoda/gradle/diffdependencytask/ConditionalDependencyEvaluatorTest.groovy
```diff
@@ -0,0 +1,104 @@
...
+import static org.mockito.Mockito.when
+import static org.mockito.MockitoAnnotations.initMocks
+
+class ConditionalDependencyEvaluatorTest {
+
+    @Mock
+    private ConditionalDependencyRepository dependenciesRepository
+    @Mock
+    private GroovyCallable<List<String>> changedFilesProvider
+    @InjectMocks
```
jackSzm: I'd rather instantiate it in the setup. `InjectMocks` are evil 👿

frapontillo: 😭

frapontillo: Done in 749db540dcc561694d5c2206e78741aa7b1fcd91

jackSzm: Main problem with `InjectMocks` is when you change c'tor and add extra parameter, `InjectMocks` doesn't throw an exception or notify you about missing parameter but simply injects a `null` there. 💥 💣 💥

PR #248, opened 2017-01-04
Android background beacon demo

ouchadam:
Demo to showcase an android application which manually handles scanning for owned beacon messages/attachments.

This demo shows how to interact with the `Google Nearby Api` in order to receive updates in the background.

A similar scenario can also be achieved without any code or the need to use a long running service by making use of the Nearby Notification functionality of `google beacon`.


android-beacon/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Mon Sep 05 16:55:21 BST 2016
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
```
blundell: I read a tweet about gradle 3.1 this morning ...

ouchadam: android studio defaults ¯\_(ツ)_/¯ will update

android-beacon/README.md
```diff
@@ -0,0 +1,23 @@
...
+## Configure your app to receive messages via the Google Nearby services
+
+https://developers.google.com/nearby/messages/android/get-started
+
+Attachments data format:
+https://developers.google.com/nearby/notifications/attachment_data_format#free-form_app_intent 
+
+## Configure secrets file
+
+In order to be able to use the project, copy the `secrets.properties.example` file to `secret.properties` and fill in the Google Nearby API key.
```
danybony: Watch out for that `s`!
Is it `secrets.properties` or `secret.properties`?
We need to update one of the two in the readme

android-beacon/src/main/java/com/novoda/beacon/MainActivity.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class MainActivity extends AppCompatActivity {
+
+    public static final int RESOLVE_PERMISSIONS = 100;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_main);
+        startService(new Intent(MainActivity.this, MessageService.class));
```
danybony: Extract this into a method, given that it's called twice

android-beacon/src/main/java/com/novoda/beacon/MessageReceiver.java
```diff
@@ -0,0 +1,58 @@
...
+    public void onReceive(Context context, Intent intent) {
+        notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+        this.context = context;
+        Nearby.Messages.handleIntent(intent, messageListener);
+    }
+
+    private final MessageListener messageListener = new MessageListener() {
+        @Override
+        public void onFound(Message message) {
+            String content = new String(message.getContent());
```
danybony: Is `message.getContent()` containing the attachment?

android-beacon/src/main/java/com/novoda/beacon/MessageService.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private final GoogleApiClient.ConnectionCallbacks connectionCallbacks = new GoogleApiClient.ConnectionCallbacks() {
+        @Override
+        public void onConnected(@Nullable Bundle bundle) {
+            Log.d(TAG, "onConnectionConnected");
+            subscribe(googleApiClient);
+        }
+
+        private void subscribe(GoogleApiClient mGoogleApiClient) {
```
danybony: :nauseated_face:  `googleApiClient `

ouchadam: oops!

android-beacon/src/main/java/com/novoda/beacon/MessageService.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Nullable
+    @Override
+    public IBinder onBind(Intent intent) {
+        return null;
+    }
+
+    @Override
+    public int onStartCommand(Intent intent, int flags, int startId) {
+        googleApiClient = new GoogleApiClient.Builder(this)
```
fourlastor: this should be in `onCreate`

android-beacon/src/main/java/com/novoda/beacon/MessageService.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private final GoogleApiClient.ConnectionCallbacks connectionCallbacks = new GoogleApiClient.ConnectionCallbacks() {
+        @Override
+        public void onConnected(@Nullable Bundle bundle) {
+            Log.d(TAG, "onConnectionConnected");
+            subscribe(googleApiClient);
+        }
+
+        private void subscribe(GoogleApiClient googleApiClient) {
```
fourlastor: should this method (and the methods used inside) go in `MessageService` instead of this callback?

android-beacon/src/main/java/com/novoda/beacon/MessageService.java
```diff
@@ -0,0 +1,90 @@
...
+    private final GoogleApiClient.OnConnectionFailedListener onConnectionFailedListener = new GoogleApiClient.OnConnectionFailedListener() {
+        @Override
+        public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
+            Log.d(TAG, "onConnectionFailed");
+            attemptToResolve(connectionResult);
+        }
+
+        private void attemptToResolve(@NonNull ConnectionResult connectionResult) {
+            try {
+                connectionResult.getResolution().send(MessageService.this, MainActivity.RESOLVE_PERMISSIONS, null);
```
frapontillo: What does this do exactly?

android-beacon/src/main/java/com/novoda/beacon/MessageReceiver.java
```diff
@@ -0,0 +1,58 @@
...
+            Notification notification = createNotification(context, content);
```
frapontillo: So in theory we could react to beacon messages by making a network request and only later showing a user-tailored notification?

android-beacon/src/main/res/layout/activity_beacon_result.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+  android:id="@+id/activity_main"
+  android:layout_width="match_parent"
+  android:layout_height="match_parent"
+  android:paddingBottom="@dimen/activity_vertical_margin"
+  android:paddingLeft="@dimen/activity_horizontal_margin"
+  android:paddingRight="@dimen/activity_horizontal_margin"
+  android:paddingTop="@dimen/activity_vertical_margin">
+  <TextView
```
frapontillo: Blank line before and after the `TextView`?

android-beacon/src/main/java/com/novoda/beacon/MessageReceiver.java
ouchadam: yep we could do anything we wanted, but having a long running service isn't the best experience for a user and we should ideally be using the `NearbyNotifications` instead which would require the user to explicitly click the notification first

android-beacon/src/main/java/com/novoda/beacon/MessageService.java
danybony: So, basically when a `ConnectionResult` fails, it can be resolved by the user (for example when missing a permission, or requiring Play Services update).
`getResolution()` returns a PendingIntent to the activity responsible of the error resolution: the Play Store for an update, a dialog for the notification to be displayed to the user, etc.
So calling `that resolution PendingIntent.send()` is gonna open an activity for result with the Result code we pass, and the result of that activity is caught by our activity, which restarts the service

android-beacon/src/main/java/com/novoda/beacon/MessageReceiver.java
frapontillo: Or we could trigger the service when the geofence is "close enough", i.e. within 1 km.

android-beacon/src/main/java/com/novoda/beacon/MessageService.java
danybony: What you comment reminds me, is that `connectionResult.getResolution()` can return null! ⚠️
@ouchadam please check for `connectionResult.hasResolution()` before that

android-beacon/src/main/java/com/novoda/beacon/MessageService.java
ouchadam: I'm easy, when the methods are inside the anon impls they're easier to extract into a class and it shows they're not _realllyyy_ part of the outer

fourlastor:
:shipit:

PR #251, opened 2017-01-11
[monkey-runner] Update android command plugin

tasomaniac:
Updated command plugin version to 1.7.0

Made the necessary changes to the setup. Since we are creating new tasks for every step, it needs more now. Hopefully in the future, the command plugin will be flexible enough so that instead of creating tasks, configuring it would be enough to do this.

### DIFF
--- a/monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
+++ b/monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
@@ -33,7 +33,7 @@ public class MonkeyConfigurationPlugin implements Plugin<Project> {
     private void configureTask(Project project, MonkeyRunnerExtension extension) {
         def runMonkeyAllTask = project.task(TASK_NAME)
 
-        def android = project.extensions.findByName("android")
+        def android = project.android
         android.command.devices().eachWithIndex { device, index ->
 
             def monkeyTask = project.task("runMonkeyDevice${index}", type: TargetedMonkey) {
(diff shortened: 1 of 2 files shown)


monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
```diff
@@ -33,30 +33,33 @@ public class MonkeyConfigurationPlugin implements Plugin<Project> {
...
         def runMonkeyAllTask = project.task(TASK_NAME)
 
-        def android = project.extensions.findByName("android")
+        def android = project.android
         android.command.devices().eachWithIndex { device, index ->
 
             def monkeyTask = project.task("runMonkeyDevice${index}", type: TargetedMonkey) {
                 packageName = extension.packageNameFilter
-                events = extension.eventsCount
+                adb = android.command.adb
```
xrigau: is this needed? having the same line everywhere seems a bit redundant, why isn't it done from the plugin itself (as a default value)?

tasomaniac: So, this was discussed with @devisnik 
We want `AdbTask` to be self-contained. It should be able to work on its own. 
When the plugin create the tasks, it of course configures the tasks and this is not necessary. 

But here, we are creating a new task with the type of `TargetedMonkey`. This created task is out of reach of the plugin. 

Before the task itself was grabbing the plugin. 

In the future, we are planning to have more customizations so that creating tasks manually would not be necessary.

tasomaniac: Or we were thinking about having a utility method in the plugin to attach `null` fields in the task. We haven't done that yet.

mr-archano:
@xrigau are we happy to ship this how it is? Improvements on the command plugin are coming soon anyway.

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
xrigau: 👍

xrigau:
yeahhhh :shipit:

PR #253, opened 2017-01-12
[monkey-trap] Fix NullPointerException caused by Service recreation

tasomaniac:
This PR fixes a crash where the service is re-created with a null Intent. 

This is the default behavior when the resources are scarce (e.g someone plays Pokemon Go) on a device.

Yes, I still play and whenever I open the game, I see MonkeyTrap crashes. 

Our service was long-living which is not really necessary when the overlay is not shown.

Now we add the view only when we have the show command.
We stop the Service completely when we have the hide command. This also removes the overlay automatically.

### DIFF
--- a/MonkeyTrap/app/build.gradle
+++ b/MonkeyTrap/app/build.gradle
@@ -2,14 +2,16 @@ apply plugin: 'com.android.application'
 
 android {
   compileSdkVersion 25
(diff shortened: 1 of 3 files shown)


florianmski:
I added @xrigau because he wrote this so he might have more insights.
LGTM otherwise 👍

PR #254, opened 2017-01-13
PT-380: [NNP] gradle-nonnull-plugin 

tobiasheine:
[PT-380](https://novoda.atlassian.net/browse/PT-380)

**Description**

This adds a gradle plugin which helps the user to enforce `@NonNull` method parameters among a whole project.

**Details**

The plugin generates [package annotations](https://blogs.oracle.com/darcy/entry/javadoc_tip_prefer_package_info) containing `@ParametersAreNonnullByDefault` for each `package` that has a file.

In the current version it just works for `android` modules, but we're planning to enable it for `java` modules as one of the next steps.

**Tests**

We are using Gradle TestKit to build the sample project and assert that the necessary files are created containing the annotation.

**Screenshots**

screen shot 2017-01-13 at 11 17 37

Paired with @tasomaniac


tasomaniac:
@mr-archano you're right. I just merged the first skeleton commits to master. It is now much better.

gradle-nonnull-plugin/plugin/src/main/groovy/com.novoda.gradle.nonnull/GeneratePackageAnnotationsTask.groovy
```diff
@@ -0,0 +1,71 @@
...
+
+class GeneratePackageAnnotationsTask extends DefaultTask {
+
+    def outputDir
+
+    def variant
+
+    @TaskAction
+    void generatePackageAnnotations() {
+        description = "Generate package-info.java classes"
```
devisnik: This could be a bit more specific, imho. 
May be `Annotates the source packages with @ParametersAreNonnullByDefault` ?

gradle-nonnull-plugin/plugin/src/main/groovy/com.novoda.gradle.nonnull/GeneratePackageAnnotationsTask.groovy
```diff
@@ -0,0 +1,71 @@
...
+
+                def file = new File(dir, 'package-info.java')
+                if (file.createNewFile()) {
+                    file.write(getFileContent(packagePath))
+                }
+            }
+        }
+    }
+
+    static def getFileContent(String path) {
```
devisnik: Or `createAnnotationDefinition` ?

gradle-nonnull-plugin/plugin/src/main/groovy/com.novoda.gradle.nonnull/GeneratePackageAnnotationsTask.groovy
```diff
@@ -0,0 +1,71 @@
...
+        def packageName = path.replaceFirst('/', '').replaceAll('/', '.')
```
devisnik: I'd suggest passing the package into the method, doing the conversion outside.

gradle-nonnull-plugin/sample/.gitignore
```diff
@@ -7,3 +7,4 @@
 /build
 /captures
 .externalNativeBuild
+package-info.java
```
devisnik: What's this for?

gradle-nonnull-plugin/sample/app/src/custom/java/com/novoda/gradle/nonnull/custom/Foo.java
```diff
@@ -0,0 +1,6 @@
+package com.novoda.gradle.nonnull.custom;
+
+public class Foo {
+
+
+}
```
devisnik: You may want to align on the number of empty lines in theses files.

gradle-nonnull-plugin/sample/app/src/main/AndroidManifest.xml
```diff
@@ -4,10 +4,8 @@
 
   <application
     android:allowBackup="true"
-    android:icon="@mipmap/ic_launcher"
     android:label="@string/app_name"
-    android:supportsRtl="true"
-    android:theme="@style/AppTheme">
+    android:supportsRtl="true">
```
devisnik: 👀

gradle-nonnull-plugin/sample/.gitignore
tasomaniac: That's how it was done before because we didn't know if it works through generated folder. It was in the sources. Should be removed now.

gradle-nonnull-plugin/plugin/src/main/groovy/com.novoda.gradle.nonnull/GeneratePackageAnnotationsTask.groovy
```diff
@@ -0,0 +1,71 @@
...
+            | * void setValue(@NonNull String value) {
+            | *     this.value = value;
+            | * }
+            | *
+            | */
+            |@ParametersAreNonnullByDefault
+            |package ${packageName};
+            |
+            |import javax.annotation.ParametersAreNonnullByDefault;
+            |""".stripMargin('|')
```
mr-archano: care to elaborate how this works?

gradle-nonnull-plugin/plugin/src/main/groovy/com.novoda.gradle.nonnull/GeneratePackageAnnotationsTask.groovy
```diff
@@ -0,0 +1,71 @@
...
+
+    def outputDir
+
+    def variant
+
+    @TaskAction
+    void generatePackageAnnotations() {
+        description = "Annotates the source packages with @ParametersAreNonnullByDefault"
+
+        variant.sourceSets.each {
```
mr-archano: I'm not sure if supported, but using `.all` instead of `.each` when possible would be more robust I guess. Probably worth checking this option.

gradle-nonnull-plugin/plugin/src/test/groovy/com/novoda/gradle/nonnull/AndroidProjectIntegrationTest.groovy
```diff
@@ -0,0 +1,62 @@
...
+
+public class AndroidProjectIntegrationTest {
+
+    @ClassRule
+    public static final ProjectRule PROJECT = new ProjectRule()
+
+    @Test
+    public void shouldGeneratePackageAnnotation() {
+        PROJECT.generatedSrcDirs.each {
+            def file = new File(it, 'package-info.java')
```
mr-archano: `nit:` I would suggest to try to avoid `def` unless properly needed. It's easier for the IDE to work out autocompletion & refactoring when type is explicit. It also minimise mistakes when code gets shuffled around.

gradle-nonnull-plugin/plugin/src/main/groovy/com.novoda.gradle.nonnull/GeneratePackageAnnotationsTask.groovy
tasomaniac: We already tried. Since it is a `Set`, there is no `all`

gradle-nonnull-plugin/plugin/src/main/groovy/com.novoda.gradle.nonnull/GeneratePackageAnnotationsTask.groovy
tasomaniac: To strip indentation on triple quoted Strings, we have 2 methods: `stripIndentation` and `stripMargin`.
Striping indentation removes all the whitespace, but we have some whitespace in the javadoc to preserve. So this strips the indentation before `|` including `|`

gradle-nonnull-plugin/plugin/src/test/groovy/com/novoda/gradle/nonnull/AndroidProjectIntegrationTest.groovy
tasomaniac: When the variable is used locally, the IDE autocompletion works because it infers the type, no? Especially in the tests, I thought this is more groovy because writing `File` twice is redundant.

gradle-nonnull-plugin/plugin/src/main/groovy/com.novoda.gradle.nonnull/GeneratePackageAnnotationsTask.groovy
mr-archano: TIL. Thanks @tasomaniac

gradle-nonnull-plugin/plugin/src/test/groovy/com/novoda/gradle/nonnull/AndroidProjectIntegrationTest.groovy
mr-archano: In my experience sometimes `def` breaks refactoring. I mentioned this as minor personal remark, probably more matter of taste. As rule of thumb I usually defer the use of `def` to those case where the type can be ambiguous (think of variants in app vs library).
Happy to keep this as it is. We probably need to agree on a common shared style for the next future - let's catch up with @devisnik soon.

tasomaniac: Let's do it soon. I agree to agree on a shared style. 👍

PR #257, opened 2017-01-17
[monkey-trap] Lower targetSdkVersion to prevent permission crash

tasomaniac:
Lower targetSdkVersion to prevent permission crash

This was already set to 22 in the first iteration. 
This [commit](https://github.com/novoda/spikes/commit/ed62756d680f9bf7d76b55601f764cefc5a4fc30) unfortunately increased it.

Now we have a comment to prevent that in the future. 
And also removed some unnecessary stuff :smile:


xrigau:
💯

ataulm:
My BAD.

PR #258, opened 2017-01-19
[NNP] PT-392: up to date check support

tasomaniac:
## Problem

Currently, we plugin works on java and android modules. 👍 
But the task runs in every build process. This is not something we want to have.

If no package structure has changed, we don't want to run again. 

## Solution

A field called `packages` created with annotation `@Input`
The code block that computes the packages is extracted into a `@Memoized` getter method: `getPackages`

## Tests

- A test is added that runs the project again and asserts that we get the UP-TO-DATE check
- Another test is added that adds a new file to a new package and asserts that we didn't get UP-TO-DATE check. 

Paired with @tobiasheine


tasomaniac:
The integration tests only tests the java-only `core` module
See this line in the PR: https://github.com/novoda/spikes/pull/258/files#diff-cd2de1e0d8c93526b5ee0b102ca0b326R66

We actually did it like that because building java module is slightly faster.

mr-archano:
Sorry @tasomaniac, I actually meant non-java-only modules (edit fail), just to be safe. I understand this can be considered enough tho. So I'm happy to ship it 👍

PR #260, opened 2017-01-20
Ensure Command plugin is applied

takecare:
The monkey runner plugin depends on the android command one. Currently, if you don't apply the command plugin you get a cryptic/not very helpful error message: `Error:Could not get unknown property 'command' for object of type com.android.build.gradle.AppExtension`.

By ensuring the plugin is applied first (just like what's already happening with the android plugin) we can provide a better error message.

### DIFF
--- a/monkey-runner/README.md
+++ b/monkey-runner/README.md
@@ -5,11 +5,13 @@ Provides gradle closure to configure the monkey runner.
 
 ## Description
 
-Decorates the [Android monkey](https://developer.android.com/studio/test/monkey.html) with the [monkey trap](https://github.com/novoda/spikes/tree/master/MonkeyTrap/), which is designed to block the monkey from accessing the system notifications tray (and the quick toggles).
+Decorates the [Android monkey](https://developer.android.com/studio/test/monkey.html) with the [monkey trap](https://github.com/novoda/spikes/tree/master/MonkeyTrap/),
+ which is designed to block the monkey from accessing the system notifications tray (and the quick toggles).
 
 ## Adding to project
 
-It's necessary to have the monkey trap installed on all the devices you want to use this monkey runner on:
+In case you want the monkey trap running, it's necessary to have it installed on all the devices you want to use 
+monkey runner with. You can easily ensure this with:
 
 ```bash
 wget https://raw.githubusercontent.com/novoda/spikes/master/MonkeyTrap/apk/app-debug.apk
@@ -22,7 +24,8 @@ rm app-debug.apk
 
 You can run this as part of your CI job before starting the monkey runner.
 
-In your Android module's `build.gradle`:
+In your Android module's `build.gradle`, add this plugin as a dependency, apply it and configure it. Be sure you also apply
+ the `android-command` plugin (which `monkey-runner` depends on):
 
 ```groovy
 buildscript {
@@ -41,6 +44,7 @@ buildscript {
 
 ...
 
+apply plugin: 'com.novoda.android-command'
 apply plugin: 'com.novoda.monkey-runner'
 
 ...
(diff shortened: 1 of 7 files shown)


monkey-runner/plugin/src/test/groovy/com/novoda/monkey/MonkeyConfigurationPluginTest.groovy
```diff
@@ -6,25 +6,37 @@ import org.gradle.testfixtures.ProjectBuilder
...
-        project = ProjectBuilder.builder().build()
+    void givenAndroidPluginNotApplied_whenApplyingMonkey_thenItThrowsException() {
+        project.apply plugin: 'android-command'
 
         project.apply plugin: MonkeyConfigurationPlugin
+
+    }
+
+    @Test(expected = GradleException.class)
+    void givenCommandPluginNotApplied_whenApplyingMonkey_thenItThrowsException() {
```
tasomaniac: The tests methods are mixed up.

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
```diff
@@ -24,9 +26,18 @@ public class MonkeyConfigurationPlugin implements Plugin<Project> {
...
 
     private static void ensureAndroidPluginAppliedTo(Project project) {
-        boolean missingAndroidPlugin = !project.plugins.hasPlugin('com.android.application')
-        if (missingAndroidPlugin) {
-            throw new GradleException('monkey runner plugin can only be applied after the Android plugin')
+        ensurePluginIsApplied('com.android.application', project)
+    }
+
+    private static void ensureCommandPluginAppliedTo(Project project) {
+        ensurePluginIsApplied('android-command', project)
```
tasomaniac: Our plugin can also be applied like `com.novoda.android-command`. 
I believe this will still work even the user applied the FQN (because gradle magic?)

monkey-runner/plugin/src/test/groovy/com/novoda/monkey/MonkeyConfigurationPluginTest.groovy
takecare: what do you mean?

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
takecare: think so. tested and it works with the fqn.

monkey-runner/README.md
```diff
@@ -41,6 +44,7 @@ buildscript {
 
 ...
 
+apply plugin: 'com.novoda.android-command'
```
tasomaniac: Should we add the `classpath` above for the command plugin as well?
Because it won't work without it.

monkey-runner/plugin/src/test/groovy/com/novoda/monkey/MonkeyConfigurationPluginTest.groovy
tasomaniac: 🙏  I was confused. I simply saw `com.android.application` below and thought it was testing android plugin not applied. 😄

monkey-runner/README.md
takecare: will do

takecare: actually, do we really need to add the command plugin to the android app's (sample) classpath?

tasomaniac: I expected so, how is it going to work without it? Is it because it is defined as a dependency in the plugin?

takecare: i think so

PR #262, opened 2017-01-23
[monkey-runner] Apply command plugin by default

takecare:
We can avoid having the users' of this plugin having to apply the android command plugin on their own, thus making it simpler to integrate this plugin in their projects.

Tests updated.

### DIFF
--- a/monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
+++ b/monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
@@ -12,10 +12,7 @@ class MonkeyConfigurationPlugin implements Plugin<Project> {
     void apply(Project project) {
         ensureAndroidPluginAppliedTo(project)
 
-        project.configure(project) {
-            //noinspection GroovyAssignabilityCheck
-            apply plugin: 'com.novoda.android-command'
-        }
+        project.apply plugin: 'com.novoda.android-command'
 
         MonkeyRunnerExtension extension = project.extensions.create(MonkeyRunnerExtension.NAME, MonkeyRunnerExtension)
         extension.setDefaultsForOptionalProperties()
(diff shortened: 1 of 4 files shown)


monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
```diff
@@ -11,7 +11,11 @@ class MonkeyConfigurationPlugin implements Plugin<Project> {
     @Override
     void apply(Project project) {
         ensureAndroidPluginAppliedTo(project)
-        ensureCommandPluginAppliedTo(project)
+
+        project.configure(project) {
```
tasomaniac: Can't we just do
`project.apply plugin: 'com.novoda.android-command'`

`apply` is a method in `Project` class.

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
```diff
@@ -11,7 +11,11 @@ class MonkeyConfigurationPlugin implements Plugin<Project> {
...
+            //noinspection GroovyAssignabilityCheck
+            apply plugin: 'com.novoda.android-command'
```
takecare: we could do something like
```
project.configure(project, {
            project.plugins.apply(AndroidCommandPlugin.class)
        })
```
instead (it solves the warning)
@mr-archano rfc

tasomaniac:
I think `README` should also be adjusted with respect to the changes here

monkey-runner/plugin/src/main/groovy/com/novoda/monkey/MonkeyConfigurationPlugin.groovy
takecare: yes we can! 👍

takecare:
@tasomaniac true, i always forget

takecare:
@tasomaniac i understand that point of view and spent some time thinking it over before suggesting that change. my reasoning is:

1. the command plugin is already a dependency either way and its tasks will always be bundled in;
2. requiring the user (dev) to apply it does make it obvious but that's it;
3. if we are clear enough about the dependency i think it works out in the end.

anyway, more than open to other people's suggestions/opinions

tasomaniac:
Makes sense. Thanks for the explanation. 
Regarding to point 3, we should maybe still have some words about command plugin in the Readme. What do you think?

takecare:
@tasomaniac haven't forgot about it! hopefully will be changing the readme today

PR #263, opened 2017-01-23
Tpbot/ NPE on service

Mecharyry:
#### Problem
Seeing a `NullPointerException` when a command is sent to the `MovementService`. Seems to be some confusion as to when `onServiceConnected(ComponentName name, IBinder service);` is called by the `Service`, it is called before `onStartCommand(Intent intent, int flags, int startId)` where the `Toaster` etc are created, resulting in an NPE.

#### Solution
Move creation of `Toaster` to onCreate to prevent NPE, which occurs before any of the other callbacks.


Mecharyry:
Errr no I didn't lol

Mecharyry:
ok, seems like my AS was on the wrong branch for a minute FML.

danybony:
Mmmkay you merged it already (it seems) but the change looked good 👍

PR #265, opened 2017-01-25
Tpbot/ Handler exception

Mecharyry:
#### Problem
Receiving an NPE when navigating between the `HumanActivity` and the `LandingActivity`. We are trying to remove callbacks on a `Handler` that is null.

#### Solution
Add guard against attempting to clear callbacks on a null `Handler`.

#### Additional
Stop and Start of presenting are matched to the corresponding lifecycle methods, onStop and onStart. Guard against NPEs on Presenter when detaching service and observables.

### DIFF
--- a/TelepresenceBot/core/src/main/java/com/novoda/tpbot/human/HumanPresenter.java
+++ b/TelepresenceBot/core/src/main/java/com/novoda/tpbot/human/HumanPresenter.java
@@ -23,8 +23,13 @@ class HumanPresenter {
     }
 
     void stopPresenting() {
-        observable.detachObservers();
-        humanTpService.disconnect();
+        if (observable != null) {
+            observable.detachObservers();
+        }
+
+        if (humanTpService != null) {
+            humanTpService.disconnect();
+        }
     }
 
     private class ConnectionObserver implements Observer<Result> {
--- a/TelepresenceBot/core/src/main/java/com/novoda/tpbot/human/HumanPresenter.java
+++ b/TelepresenceBot/core/src/main/java/com/novoda/tpbot/human/HumanPresenter.java
@@ -27,9 +27,7 @@ class HumanPresenter {
             observable.detachObservers();
         }
 
-        if (humanTpService != null) {
-            humanTpService.disconnect();
-        }
+        humanTpService.disconnect();
     }
 
     private class ConnectionObserver implements Observer<Result> {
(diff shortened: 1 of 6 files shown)


TelepresenceBot/core/src/main/java/com/novoda/tpbot/human/HumanPresenter.java
```diff
@@ -23,8 +23,13 @@ void startPresenting(String serverAddress) {
     }
 
     void stopPresenting() {
-        observable.detachObservers();
-        humanTpService.disconnect();
+        if (observable != null) {
```
jackSzm: Shouldn't this be DeveloperError ? When you `stopPresenting` before calling `startPresenting` ?

TelepresenceBot/core/src/main/java/com/novoda/tpbot/human/HumanPresenter.java
```diff
@@ -23,8 +23,13 @@ void startPresenting(String serverAddress) {
...
+            observable.detachObservers();
+        }
+
+        if (humanTpService != null) {
```
jackSzm: If this can be null I just wouldn't pass it in the ctor

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/human/HumanActivity.java
```diff
@@ -95,6 +81,12 @@ public void onConnect(String serverAddress) {
     };
 
     @Override
+    protected void onStart() {
+        super.onStart();
+        presenter = new HumanPresenter(SocketIOTpService.getInstance(), this);
```
jackSzm: And if we don't pass `service` in the constructor, initialisation can still be done in `onCreate` and here just some sort of a setter. Or startPresenting with a `service` as a parameter ?

TelepresenceBot/core/src/main/java/com/novoda/tpbot/human/HumanPresenter.java
Mecharyry: In this case probably not. This presenter is not bound to only the `start` and `stop` of an `activity` but to a `ViewSwitcher`.

TelepresenceBot/core/src/main/java/com/novoda/tpbot/human/HumanPresenter.java
Mecharyry: This should never be null. I didn't think about it when I did this change 😆

PR #267, opened 2017-01-25
Tpbot bot mimic human

Mecharyry:
#### Problem
We could simplify the `BotActivity` by mimicking the `HumanActivity`, we don't really need to supply a `RecyclerView` of direction items for debugging purposes. This should speed up development of the debug app.

#### Solution
Mimic the behaviour of the `HumanActivity`.


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/BotActivity.java
```diff
@@ -78,6 +78,13 @@ public void onLazersReleased() {
         }
     };
 
+    private final ServerDeclarationListener serverDeclarationListener = new ServerDeclarationListener() {
```
danybony: Wait a second, why do we need this here?
The bot _is_ the server or did I miss something?
Are we considering having a man-in-the-middle server?

Mecharyry: We have a man in the middle server. I could probably set it up so that one is the server and one is the client but I haven't looked into it yet. Would rather have this change in, as I know it works and work on having one server one client.

danybony: Ok, it's fine then, I missed that bit regarding the shared server

danybony:
LGTM, let's keep the momentum 👍 🚢

PR #268, opened 2017-01-26
Tpbot/ Bot service

Mecharyry:
#### Problem
We have a `BotPresenter` and a `BotView` but no `TpService` implementation 😱 

#### Solution 
Add a `SocketIO` implementation of `BotTpService`. Extract `SocketConnectionObservable` as both the `Human` and `Bot` connect in the same way but with slightly different events.

#### Next Steps
- Add server implementation code as a separate module in repo. 
- Move generic classes to more appropriate package. Avoided in this PR because it would mess up diff.


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/SocketIOTpService.java
```diff
@@ -0,0 +1,91 @@
...
+    @Override
+    public Observable<Direction> listen() {
+        return new DirectionObservable();
+    }
+
+    private class DirectionObservable extends Observable<Direction> {
+
+        @Override
+        public Observable<Direction> start() {
+            if (!socket.connected()) {
```
tasomaniac: Just asking, is there a possibility that this will stay disconnected? I'm asking because we are ignoring when it is not connected. Is that fine?

Mecharyry: It has automatic retry built in. I'm handling the case when we get disconnected with a `Result.error` which will make the user connect to the server again.

tasomaniac:
@danybony why u no merge dis PR

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/SocketIOTpService.java
tasomaniac: Thanks man

danybony:
I was waiting for the answer to your question @tasomaniac 😂

PR #270, opened 2017-03-07
Add ViewActivityRule to espresso-support module

ataulm:
This PR adds the ViewActivityRule to the espresso support module.

This rule allows you to pass an XML layout, specifying the type of the root, to an activity test rule.

It will inflate the view into an activity, so you can write tests for a view in isolation.

---

It's pretty dumb, the aim is to test the behaviour we have in custom views. They're similar to the BDD  style unit tests we write in that you can have given-when-then.

- 'given' is binding some data to the view
- 'when' is an action you perform on the view (e.g. tapping it, or tapping a child view)
- 'then' is the assertion made either on a view property changing or verifying a callback was fired for a given listener using Mockito

---

Please note big question: https://github.com/novoda/spikes/pull/270#issuecomment-284661870


espresso-support/demo/src/androidTest/java/com/novoda/movies/MovieItemViewTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+@RunWith(AndroidJUnit4.class)
+@LargeTest
+public class MovieItemViewTest {
+
+    private static final Movie EDWARD_SCISSORHANDS = new Movie("Edward Scissorhands");
+    private static final Movie NOT_EDWARD_SCISSORHANDS = new Movie("NOT Edward Scissorhands");
+
+    @Rule
+    public ViewActivityRule<MovieItemView> viewActivityRule = new ViewActivityRule<>(R.layout.test_movie_item_view);
```
ataulm: pass the layout, the type is the root view of the layout

espresso-support/demo/src/androidTest/java/com/novoda/movies/MovieItemViewTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private HitCountingMovieItemViewListener hitCountingListener;
```
ataulm: we'll use this (it's defined at the bottom of this test class) to check if methods on our listener were called

espresso-support/demo/src/androidTest/java/com/novoda/movies/MovieItemViewTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Before
+    public void setUp() {
+        MovieItemView view = viewActivityRule.getView();
+
+        hitCountingListener = new HitCountingMovieItemViewListener();
+        view.attachListener(hitCountingListener);
```
ataulm: If your custom view has a separate method for attaching a listener, you can attach it in the `setUp()` method of the test, since it shouldn't be data dependent.

If it's passed in to the `bind()`/`update()` method, you can pass it at the point of binding instead, no big deal.

espresso-support/demo/src/androidTest/java/com/novoda/movies/MovieItemViewTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+        // setup standard 'given' used for all tests
+        viewActivityRule.bindViewUsing(new ViewActivityRule.Binder<MovieItemView>() {
+            @Override
+            public void bind(MovieItemView view) {
+                view.bind(EDWARD_SCISSORHANDS);
```
ataulm: we pass a `Binder` to the `ViewActivityRule` because it calls the bind on the uiThread and don't want the client to have to type that every time.

Here, this is like setting up your test scenarios "given MovieItemView is displaying EdwardScissorhands"

espresso-support/demo/src/androidTest/java/com/novoda/movies/MovieItemViewTest.java
```diff
@@ -0,0 +1,140 @@
...
+    @After
+    public void tearDown() {
+        viewActivityRule.getView().detachListeners();
+    }
+
+    @Test
+    public void clickMovieItemView() {
+        onView(withClassName(is(MovieItemView.class.getName()))).perform(click());
+
+        hitCountingListener.assertClick(EDWARD_SCISSORHANDS);
```
ataulm: our first assert!

```
(givenMovieItemViewIsDisplayingEdwardScissorhands)

whenClickingOnTheMovieItemView

thenCallsClickEdwardScissorhands
```

espresso-support/demo/src/androidTest/java/com/novoda/movies/MovieItemViewTest.java
```diff
@@ -0,0 +1,140 @@
...
+        hitCountingListener.assertClickPlay(EDWARD_SCISSORHANDS);
+    }
+
+    @Test
+    public void rebindWithDifferentMovie_noClickMoviePlayView() {
+        // setup 'given' in test
+        viewActivityRule.bindViewUsing(new ViewActivityRule.Binder<MovieItemView>() {
+            @Override
+            public void bind(MovieItemView view) {
+                view.bind(NOT_EDWARD_SCISSORHANDS);
```
ataulm: you can also bind inside a test, if you need to. This could be a very common case, since the aim of these types of tests is to assert behaviour given different input - it depends on whether your views behave differently given different inputs.

e.g. if a movie was "not playable", we could test that the play icon was not displayed.

espresso-support/demo/src/androidTest/java/com/novoda/movies/MovieItemViewTest.java
```diff
@@ -0,0 +1,140 @@
...
+            }
+        });
+
+        onView(withId(R.id.movie_item_button_play)).perform(click());
+
+        // like `Mockito.verify(foo, never())`
+        hitCountingListener.assertNoClickPlay(EDWARD_SCISSORHANDS);
```
ataulm: In this particular case, we test that `clickPlay({movie("Edward Scissorhands")})` is never called.

espresso-support/demo/src/androidTest/java/com/novoda/movies/MovieItemViewTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+        private static final String ON_CLICK = "onClick ";
+        private static final String ON_CLICK_PLAY = "onClickPlay ";
+        private static final String ON_CLICK_FAVORITE = "onClickFavorite ";
+
+        private final HitCounter hitCounter = new HitCounter();
+
+        @Override
+        public void onClick(Movie movie) {
+            hitCounter.markHit(ON_CLICK + movie.name);
```
ataulm: the `HitCounter` is a wrapper around a `Map<CharSequence,Boolean>`. This means you can be specific with your asserts by generating keys as a combination of method calls and method params as we do here.

espresso-support/demo/src/androidTest/java/com/novoda/movies/MovieItemViewTest.java
```diff
@@ -0,0 +1,140 @@
...
+            }
+        });
+
+        onView(withId(R.id.movie_item_button_play)).perform(click());
+
+        // like `Mockito.verify(foo, never())`
+        hitCountingListener.assertNoClickPlay(EDWARD_SCISSORHANDS);
+    }
+
+    private static class HitCountingMovieItemViewListener implements MovieItemView.Listener {
```
ataulm: This wraps the library's `HitCounter` so we can bind it to the `MovieItemView` as a standard listener. Whenever the listener's methods are called, instead of delegating back to the activity/presenter, we'll just record "hits".

espresso-support/demo/src/debug/res/layout/test_movie_item_view.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<com.novoda.movies.MovieItemView xmlns:android="http://schemas.android.com/apk/res/android"
```
ataulm: we wanted to do it so that you didn't even have to specify a layout file. Most of the custom views at Novoda have logic in `onFinishInflate()` (e.g. dependency/collaborator instantiation) and `onFinishInflate()` will not be called unless the view was inflated from XML.

espresso-support/demo/src/main/java/com/novoda/movies/MovieItemView.java
```diff
@@ -1,52 +1,32 @@
...
-import com.novoda.accessibility.ActionsAccessibilityDelegate;
-import com.novoda.accessibility.ActionsAlertDialogCreator;
-
-import java.util.Arrays;
-import java.util.Locale;
-
-import butterknife.BindView;
-import butterknife.ButterKnife;
-
 public class MovieItemView extends RelativeLayout {
```
ataulm: The changes in this class should be in the base branch, dunno why they're showing up. It's just removing ButterKnife and Accessibilitools to make the demo simpler.

ataulm:
This message still comes up:

```
Caused by: java.lang.RuntimeException: Unable to resolve activity for: Intent { act=com.novoda.espresso.SHOW_VIEW flg=0x14000000 cmp=com.novoda.movies/com.novoda.espresso.ViewActivity (has extras) }
```

so it didn't look like it was fixed by using an Action.

ataulm:
It's because the Activity is defined inside the `lib` module and the `lib` module is included in the demo with `androidTestCompile`. This means the activity doesn't get registered with the package manager.

It can be "fixed" by using `debugCompile` or `espressoCompile` (where `espresso` is your test build type) but it'll include the espresso dependencies into your main apk. Whether that's fine in your debug apk is up to the user.

We hit this on talkback-toggle and decided to have two artifacts. In this case, the two artifacts would be:

- activity and manifest
- test rule

espresso-support/build.gradle
```diff
@@ -1,4 +1,6 @@
-apply from: 'dependencies.gradle'
+buildscript {
+    apply from: 'dependencies.gradle'
```
ataulm: ✨

espresso-support/core/build.gradle
```diff
@@ -3,33 +3,34 @@ buildscript {
...
     defaultConfig {
-        minSdkVersion androidMinSdkVersion
+        minSdkVersion versions.androidSdk.min
     }
 }
 
 dependencies {
-    compile librariesTest.androidTestRunner
-    compile librariesTest.androidEspressoCore
+    provided project(':extras')
```
ataulm: `extras` includes `ViewActivity`.

This means clients can add the `extras` artifact as `debugCompile/espressoCompile` as part of their app under test, so the activity is registered by the package manager.

The `core` artifact just needs to know about `ViewActivity` for compilation. If the client is not using `ViewActivityRule` then they don't need to include the `extras` artifact, and can use the `core` artifact.

of course, the core artifact only includes `ViewActivityRule` right now so might be useful 😁  but we can also add common espresso things in here like custom gestures.

espresso-support/demo/src/androidTest/java/com/novoda/movies/DemoMainActivityTest.java
```diff
@@ -0,0 +1,24 @@
...
+@RunWith(AndroidJUnit4.class)
+@LargeTest
+public class DemoMainActivityTest {
+
+    @Rule
+    public ActivityTestRule<DemoMainActivity> activityRule = new ActivityTestRule<>(DemoMainActivity.class);
+
+    @Test
+    public void testActivityNotNull() {
+        if (activityRule.getActivity() == null) {
```
ataulm: this test is just to assert that the `extras` artifact is not necessary if you're not using `ViewActivityRule`. If you comment out that dependency, then this test will run perfectly fine. The MovieItemViewTest will _compile_ fine, but will fail at runtime because it cannot resolve `ViewActivity` which is in the `extras` artifact.

espresso-support/core/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1 @@
+<manifest package="com.novoda.espresso.core" />
```
rock3r: The most barebones manifest ever :D

espresso-support/demo/build.gradle
```diff
@@ -36,11 +36,17 @@ android {
 }
 
 dependencies {
-    androidTestCompile project(':lib')
-    androidTestCompile(librariesTest.androidTestRunner) {
+    debugCompile project(':extras')
```
rock3r: How come this is still for `debugCompile`?

espresso-support/demo/src/main/AndroidManifest.xml
```diff
@@ -7,6 +7,10 @@
         android:icon="@drawable/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/AppTheme"
-        tools:ignore="GoogleAppIndexingWarning" />
+        tools:ignore="GoogleAppIndexingWarning">
+
+        <activity android:name=".DemoMainActivity" />
```
rock3r: Fully qualified name?

espresso-support/dependencies.gradle
```diff
@@ -1,23 +1,25 @@
...
-
-    librariesBuild = [
-        androidGradleBuildTools: 'com.android.tools.build:gradle:2.3.0',
-        bintrayRelease         : 'com.novoda:bintray-release:0.4.0'
+    versions = [
+            androidSdk    : [
+                    compile   : 25,
+                    min       : 16,
+                    target    : 25,
+                    buildTools: '25.0.1'
```
rock3r: Can we set this to the latest (25.0.2) while we're at it? Same for support lib, 25.2.0 (critical bugfix from 25.0 and 25.1)

espresso-support/dependencies.gradle
```diff
@@ -1,23 +1,25 @@
...
+            build              : [
+                    androidGradle : 'com.android.tools.build:gradle:2.3.0',
+                    bintrayRelease: 'com.novoda:bintray-release:0.4.0'
+            ],
+            androidTestRunner  : 'com.android.support.test:runner:0.5',
+            androidEspressoCore: 'com.android.support.test.espresso:espresso-core:2.2.2',
+            dexmaker           : 'com.google.dexmaker:dexmaker:1.2',
+            dexmakerMockito    : 'com.google.dexmaker:dexmaker-mockito:1.2',
+            jUnit              : 'junit:junit:4.12',
+            mockitoCore        : 'org.mockito:mockito-core:1.10.19',
```
rock3r: Why not mockito 2? You will also need a [more recent version of dexmaker](https://github.com/linkedin/dexmaker/blob/master/CHANGELOG.md#version-220-2016-12-09) if you upgrade btw

espresso-support/demo/src/main/AndroidManifest.xml
ataulm: I'm not sure we do want this anymore - fully qualified seems to be a remnant of old eclipse/early AS. AS is happy with refactoring `.Foo` now.

espresso-support/demo/build.gradle
ataulm: This contains the ViewActivity so it needs to be compiled into the apk under test. This might be debug sourceset or you might have a specific sourceset/build flavor for espresso.

The `core` artifact is the one that's explicitly only needed for `androidTestCompile`

espresso-support/dependencies.gradle
ataulm: yiss

espresso-support/demo/src/main/AndroidManifest.xml
rock3r: Ok, I'm cool with that

espresso-support/dependencies.gradle
ataulm: This is just what I copy and paste from whatever is handy - if it works I'm happy. It's not being bundled into the artifacts so I don't mind leaving as is, but will change since i will change the other dependencies.

espresso-support/dependencies.gradle
ataulm: commit 9425c9a537cb71d8239eb0e1ff392ce7438ecf13
Author: Ataul Munim <hello@ataulm.com>
Date:   Fri Mar 10 16:08:14 2017 +0000

    Update android build tools and support library version

espresso-support/dependencies.gradle
ataulm: I didn't change this one because I would have had to switch to a different set of dependencies (dexmaker from linkedin or someone else) instead of just the bump of google's ones.

PR #271, opened 2017-03-11
Tpbot/ Distinguish between Humans and Bots

Mecharyry:
#### Problem
We have a middleman server but it does not currently distinguish between a bot and a human.

#### Solution
Add server code to distinguish between bots and humans, only allowing a human to connect if a bot is already in a room. Simplify the client logic so that rather than sending custom events through socket io we use the built-in events such as `connect`, `error` and `error_connect`.

#### Follow-up PR
Will forward `movement` events from the `human` client to the `bot` client.


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/SocketIOTpService.java
```diff
@@ -35,12 +36,14 @@ private SocketIOTpService() {
     public Observable<Result> connectTo(String serverAddress) {
         try {
             URL url = new URL(serverAddress);
-            socket = IO.socket(url.toExternalForm());
+            IO.Options options = new IO.Options();
+            options.query = ClientType.BOT.rawClientType();
```
Mecharyry: Pass a query string rather than creating separate events for the human and bot connection.

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/ClientType.java
```diff
@@ -0,0 +1,18 @@
+package com.novoda.tpbot.support;
+
+public enum ClientType {
```
Mecharyry: This is used on the server to distinguish between a `human` and a `bot` client.

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/SocketConnectionObservable.java
```diff
@@ -4,60 +4,65 @@
...
-        socket.emit(event.rawEvent(), "", joinSuccessAcknowledgementListener);
         socket.on(EVENT_CONNECT_ERROR, connectionTimeoutListener);
+        socket.on(EVENT_ERROR, connectionErrorListener);
+        socket.on(EVENT_CONNECT, connectionEstablishedListener);
         socket.connect();
         return this;
     }
 
-    private final Ack joinSuccessAcknowledgementListener = new Ack() {
+    private final Emitter.Listener connectionTimeoutListener = new Emitter.Listener() {
```
Mecharyry: Remove custom `connection` event in favour of using a basic listener that is tied to already built-in socket io events.

TelepresenceBot/server-unit/core/loggingClient.js
```diff
@@ -0,0 +1,14 @@
+function LoggingClient(client) {
+    this.client = client;
+}
+
+LoggingClient.prototype.emit = function(event, dataToEmit) {
```
Mecharyry: This function mirrors the socket io `client` so that for real runs of the server we use this logging client and for testing we can use an actual socket io `client`.

TelepresenceBot/server-unit/core/server.js
```diff
@@ -1,21 +1,64 @@
...
+var ClientType = require("./clientType.js");
+var LoggingClient = require("./loggingClient.js");
 
-var clients = {};
+var room = "London";
+var humans = [];
+var bots = [];
+var testClient = new LoggingClient();
+
+io.use(function(client, next){
```
Mecharyry: This function establishes whether the client is allowed to connect to the server.

A `Human` client is allowed to connect when a bot is present in the `London` room and no other `Human` client is present in the same room.

A `Bot` and a `Test` client is always allowed to connect regardless.

TelepresenceBot/server-unit/core/server.js
```diff
@@ -1,21 +1,64 @@
...
 io.sockets.on('connection', function (client) {
 
-    console.log('connecting: ' + client.id);
-    clients[client.id] = client;
-    io.sockets.emit('connected', toKeysArrayFrom(clients));
+    var rawClientType = client.handshake.query.clientType;
+    var clientType = ClientType.from(rawClientType);
+
+    switch(clientType) {
+        case ClientType.HUMAN:
```
Mecharyry: `Human`s and `Bot`s get added to their respective arrays and then join the `London` room.

TelepresenceBot/server-unit/core/server.js
```diff
@@ -1,21 +1,64 @@
...
+
 
     client.on('disconnect', function() {
-        console.log('disconnecting: ' + client.id);
-        delete clients[client.id];
-        io.sockets.emit('disconnected', toKeysArrayFrom(clients));
+        humans.splice(humans.indexOf(client.id), 1);
+        bots.splice(bots.indexOf(client.id), 1);
+        testClient.emit('disconnected_human', humans);
+        testClient.emit('disconnected_bot', bots);
```
danybony: Shouldn't you do the check for `clientType` here as well?

Mecharyry: I could do the check here. For the splice they should be in the correct list, if they are not then the tests will pick up on it which is why the `testClient` requests both of the lists whenever a client is disconnected rather than just the one it is interested in.

danybony: Ok, I see, good then 👍

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/SocketConnectionObservable.java
```diff
@@ -4,60 +4,65 @@
 
 import com.novoda.tpbot.Result;
 
-import io.socket.client.Ack;
 import io.socket.client.Socket;
 import io.socket.client.SocketIOException;
 import io.socket.emitter.Emitter;
 
-import static io.socket.client.Socket.EVENT_CONNECT_ERROR;
+import static io.socket.client.Socket.*;
```
alexstyl: Why `*`?

Mecharyry: Ah, this happens when you have more than three imports from the same package, it collapses when auto-formatting. I'll try and put them back in.

Mecharyry: done 37539cb

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/ClientType.java
```diff
@@ -0,0 +1,18 @@
...
+
+    BOT("clientType=bot"),
+    HUMAN("clientType=human");
+
+    private final String rawClientType;
```
Mecharyry: done 58fe12c

PR #272, opened 2017-03-12
Tpbot forward moves

Mecharyry:
#### Problem
A `Bot` and a `Human` client can connect to each other but currently, no communication takes place.

#### Solution
Forward movement events from the `Human` client to the `Bot` client.

#### Screen Capture
tpbot mov

### DIFF
--- a/TelepresenceBot/core/src/main/java/com/novoda/tpbot/bot/BotPresenter.java
+++ b/TelepresenceBot/core/src/main/java/com/novoda/tpbot/bot/BotPresenter.java
@@ -30,6 +30,7 @@ class BotPresenter {
         unsubscribe(connectionObservable);
         unsubscribe(directionObservable);
         tpService.disconnect();
+        botView.onDisconnect();
     }
 
     void startListeningForDirection() {
--- a/TelepresenceBot/core/src/main/java/com/novoda/tpbot/bot/BotPresenter.java
+++ b/TelepresenceBot/core/src/main/java/com/novoda/tpbot/bot/BotPresenter.java
@@ -33,12 +33,6 @@ class BotPresenter {
         botView.onDisconnect();
     }
 
-    void onConnect() {
-        directionObservable = tpService.listen()
-                .attach(new DirectionObserver())
-                .start();
-    }
-
     private class ConnectionObserver implements Observer<Result> {
 
         @Override
@@ -47,6 +41,10 @@ class BotPresenter {
                 botView.onError(result.exception().get().getMessage());
             } else {
                 botView.onConnect(result.message().get());
+
+                directionObservable = tpService.listen()
+                        .attach(new DirectionObserver())
+                        .start();
             }
         }
 
--- a/TelepresenceBot/core/src/main/java/com/novoda/tpbot/bot/BotPresenter.java
+++ b/TelepresenceBot/core/src/main/java/com/novoda/tpbot/bot/BotPresenter.java
@@ -33,7 +33,7 @@ class BotPresenter {
         botView.onDisconnect();
     }
 
-    void startListeningForDirection() {
+    void onConnect() {
         directionObservable = tpService.listen()
                 .attach(new DirectionObserver())
                 .start();
--- a/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/BotActivity.java
+++ b/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/BotActivity.java
@@ -30,6 +30,7 @@ import java.util.HashMap;
 
 public class BotActivity extends AppCompatActivity implements BotView {
 
+    private static final String EMPTY = "";
     private SelfDestructingMessageView debugView;
     private SwitchableView switchableView;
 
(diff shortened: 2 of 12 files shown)


TelepresenceBot/core/src/main/java/com/novoda/tpbot/bot/BotPresenter.java
```diff
@@ -30,6 +30,7 @@ void stopPresenting() {
         unsubscribe(connectionObservable);
         unsubscribe(directionObservable);
         tpService.disconnect();
+        botView.onDisconnect();
```
Mecharyry: When disconnecting locally, performing a stop of the activity, we need to reset the views.

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/BotActivity.java
```diff
@@ -189,7 +196,7 @@ public void onError(String message) {
 
     @Override
     public void moveIn(Direction direction) {
-        debugView.showPermanently(direction.visualRepresentation());
+        debugView.showTimed(direction.visualRepresentation());
```
Mecharyry: To match that of the human otherwise we won't know when events are no longer being emitted.

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/human/HumanActivity.java
```diff
@@ -66,31 +67,32 @@ public void onDirectionReleased(Direction direction) {
 
         @Override
         public void onLazersFired() {
-            commandRepeater.startRepeatingCommand(LAZERS);
+            // no-op for debug
```
Mecharyry: Removed the command because we are only sending `Direction` which this is not a part of.

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/SocketConnectionObservable.java
```diff
@@ -68,4 +69,16 @@ public void run() {
         }
     };
 
+    private final Emitter.Listener disconnectionListener = new Emitter.Listener() {
```
Mecharyry: When a `Bot` or `Human` is disconnected it disconnects whoever it is communicating with too. This is to ensure that the UI will be updated in the event that a `disconnection` event occurs.

TelepresenceBot/server-unit/core/loggingClient.js
```diff
@@ -6,7 +6,11 @@ LoggingClient.prototype.emit = function(event, dataToEmit) {
     if(this.client != undefined) {
         this.client.emit(event, dataToEmit);
     }
-    console.log("\nEvent Emitted: " + event + "\nData Emitted: " + dataToEmit + "\n");
+    console.log("\nEvent Emitted: " + event);
+
+    if(dataToEmit != undefined && dataToEmit.length > 0) {
```
Mecharyry: Just split this up a little bit because sometimes I would log when we actually have no data to log 😄

TelepresenceBot/server-unit/core/server.js
```diff
@@ -36,11 +36,13 @@ io.sockets.on('connection', function (client) {
     switch(clientType) {
         case ClientType.HUMAN:
             humans.push(client.id);
+            leaveAllRooms(client);
```
Mecharyry: By default, all clients are in their own room matching their client id. To ship quickly I am limiting us to one room for testing purposes, hence removing them from all rooms and adding them to the only room 'London'. Practically this means that we only have one `bot` and one `human` capable of connecting atm.

In a follow-up PR I'll actually iterate through all of the bots, checking their rooms to see if they are connected with a human and if not allowing the connection to go ahead with a new human.

TelepresenceBot/server-unit/core/server.js
```diff
@@ -55,10 +57,40 @@ io.sockets.on('connection', function (client) {
...
+    });
+
+    function leaveAllRooms(client) {
+        var rooms = Object.keys(io.sockets.adapter.sids[client.id]);
+        for(var i = 0; i < rooms.length; i++) {
+            client.leave(rooms[i]);
+        }
+    }
+
+    function disconnectRoom(name) {
```
Mecharyry: When one connection drops, either the bot or human, then all other connections in that room are also disconnected.

TelepresenceBot/server-unit/test/botTest.js
```diff
@@ -56,4 +62,30 @@ describe("TelepresenceBot Server: BotTest ",function() {
         });
     });
 
+    it('Should forward movement directions from human to bot.', function(done) {
```
Mecharyry: MOAR tests 😆

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/BotActivity.java
```diff
@@ -177,8 +177,15 @@ public void onServiceDisconnected(ComponentName componentName) {
 
     @Override
     public void onConnect(String message) {
-        debugView.showPermanently(getResources().getString(R.string.connected));
+        debugView.showPermanently(getString(R.string.connected));
         switchableView.setDisplayedChild(1);
+        presenter.startListeningForDirection();
```
takecare: i'd maybe just "forward" the `onConnect` to the presenter instead

Mecharyry: Wondering if I shouldn't drop this from the activity and just have the presenter automatically start listening to moves on connection?

takecare: makes sense. the more we can offload from the activity, the better.

PR #275, opened 2017-03-23
Tpbot/ Pass room in Query

Mecharyry:
#### Problem
Currently, the `server` only allows a single `bot` and `human` to connect to the sole `London` room. Ideally, we should pass the room in via the request query and have that `human` / `bot` join that room.

#### Solution
Move the decision of rooms to the `android app` and pass which room to join via the request query.

##### When connecting 
For Bots, they immediately join whichever room is passed in through the `query`, at the moment this will be `London`.

For Humans, pull all `sockets` located in the room passed through in `query`. Take all of the rooms that each socket belongs to, should be `London` and `socket.id` rooms. If the `socket.id` room for a socket is empty then that `socket` is not connected to another client and is our available bot. Before allowing the `human` to connect we modify the `query` sent to change the room from `London` to the `availableBotSocket.id`. On connection, the `human` is connected directly to the `bot` via the `bots` own room.

##### When disconnecting
For Bots, we disconnect all other sockets that are in the `botSocket.id` room, disconnecting all humans.

For Humans, we disconnect only the `human` socket and leave all bots in limbo, awaiting a new `human` connection.

#### Screen Capture
UI has not changed but you can now add multiple bots to a given room.

#### Future Plans
I would like to remove the `Room` from the app entirely and instead serve it to the app through a call when first opening the app.

### DIFF
--- a/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Room.java
+++ b/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Room.java
@@ -1,0 +1,17 @@
+package com.novoda.tpbot.support;
+
+public enum Room {
+
+    LONDON("room=London");
+
+    private final String rawQuery;
+
+    Room(String rawQuery) {
+        this.rawQuery = rawQuery;
+    }
+
+    public String rawQuery() {
+        return rawQuery;
+    }
+
+}
--- a/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/SocketConnectionObservable.java
+++ b/TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/SocketConnectionObservable.java
@@ -8,9 +8,7 @@ import io.socket.client.Socket;
 import io.socket.client.SocketIOException;
 import io.socket.emitter.Emitter;
 
-import static io.socket.client.Socket.EVENT_CONNECT;
-import static io.socket.client.Socket.EVENT_CONNECT_ERROR;
-import static io.socket.client.Socket.EVENT_ERROR;
+import static io.socket.client.Socket.*;
 
 public class SocketConnectionObservable extends Observable<Result> {
 
--- a/TelepresenceBot/server-unit/core/botLocator.js
+++ b/TelepresenceBot/server-unit/core/botLocator.js
@@ -5,16 +5,17 @@ function BotLocator(rooms) {
 BotLocator.prototype.locateFirstAvailableBotIn = function(room) {
     var botsInRoom = this.rooms[room];
 
-    if(botsInRoom != undefined) {
-        for (socketId in botsInRoom.sockets) {
-            var socketsInBotRoom = this.rooms[socketId];
+    if(!botsInRoom) {
+        return;
+    }
 
-            if(botNotConnectedToHuman(socketsInBotRoom)) {
-                return socketId;
-            }
+    for (socketId in botsInRoom.sockets) {
+        var socketsInBotRoom = this.rooms[socketId];
+
+        if(botNotConnectedToHuman(socketsInBotRoom)) {
+            return socketId;
         }
     }
-    return undefined;
 }
 
 function botNotConnectedToHuman(socketsInBotRoom) {
--- a/TelepresenceBot/server-unit/core/botLocator.js
+++ b/TelepresenceBot/server-unit/core/botLocator.js
@@ -1,0 +1,27 @@
+function BotLocator(rooms) {
+    this.rooms = rooms;
+}
+
+BotLocator.prototype.locateFirstAvailableBotIn = function(room) {
+    var botsInRoom = this.rooms[room];
+
+    if(botsInRoom != undefined) {
+        for (socketId in botsInRoom.sockets) {
+
+            var socketsInBotRoom = this.rooms[socketId];
+
+            if(botNotConnectedToHuman(socketsInBotRoom)) {
+                return socketId;
+            }
+        }
+    }
+    return undefined;
+}
+
+function botNotConnectedToHuman(socketsInBotRoom) {
+    return  socketsInBotRoom.length != undefined && socketsInBotRoom.length == 1;
+}
+
+module.exports = BotLocator;
+
+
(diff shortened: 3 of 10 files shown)


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/Room.java
```diff
@@ -0,0 +1,17 @@
+package com.novoda.tpbot.support;
+
+public enum Room {
```
Mecharyry: I want to remove this soon with another call to the server to fetch a collection of `Rooms` that the app can use. That way, if we want to add rooms later we don't need to redeploy the app.

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/support/SocketConnectionObservable.java
```diff
@@ -8,9 +8,7 @@
 import io.socket.client.SocketIOException;
 import io.socket.emitter.Emitter;
 
-import static io.socket.client.Socket.EVENT_CONNECT;
-import static io.socket.client.Socket.EVENT_CONNECT_ERROR;
-import static io.socket.client.Socket.EVENT_ERROR;
+import static io.socket.client.Socket.*;
```
Mecharyry: 😡

Mecharyry: Now it has four it constantly updates it automatically, I'll see if I can disable later.

TelepresenceBot/server-unit/core/loggingClient.js
```diff
@@ -6,11 +6,7 @@ LoggingClient.prototype.emit = function(event, dataToEmit) {
     if(this.client != undefined) {
         this.client.emit(event, dataToEmit);
     }
-    console.log("\nEvent Emitted: " + event);
-
-    if(dataToEmit != undefined && dataToEmit.length > 0) {
-        console.log("Data Emitted: " + dataToEmit);
-    }
+    console.log("\nEvent Emitted: ", event, "\nData Emitted: ", dataToEmit, "\n");
```
Mecharyry: Urgh fail. I want the other version.

TelepresenceBot/server-unit/core/server.js
```diff
@@ -72,25 +71,29 @@ io.sockets.on('connection', function (client) {
...
+        for(var client in clients) {
+            var connectedClient = io.sockets.connected[client];
+            connectedClient.disconnect();
         }
     }
+}
 
-    function disconnectRoom(name) {
-        var room = io.sockets.adapter.rooms[name];
+function asRoomsWithSocketIds() {
```
Mecharyry: The tests now output an object with property keys that match a room name and values that match the sockets in the room.

TelepresenceBot/server-unit/core/botLocator.js
```diff
@@ -0,0 +1,26 @@
+function BotLocator(rooms) {
+    this.rooms = rooms;
+}
+
+BotLocator.prototype.locateFirstAvailableBotIn = function(room) {
+    var botsInRoom = this.rooms[room];
+
+    if(botsInRoom != undefined) {
```
takecare:
```
if (!botsInRoom) {
  return;
}

(...)
```

TelepresenceBot/server-unit/core/botLocator.js
```diff
@@ -0,0 +1,26 @@
...
+            if(botNotConnectedToHuman(socketsInBotRoom)) {
+                return socketId;
+            }
+        }
+    }
+    return undefined;
+}
+
+function botNotConnectedToHuman(socketsInBotRoom) {
+    return  socketsInBotRoom.length != undefined && socketsInBotRoom.length == 1;
```
takecare: `return socketsInBotRoom.length === 1`?

TelepresenceBot/server-unit/core/botLocator.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+module.exports = BotLocator;
+
+
```
takecare: 👾

TelepresenceBot/server-unit/core/server.js
```diff
@@ -16,53 +17,51 @@ io.use(function(client, next){
         case ClientType.BOT:
             return next();
         case ClientType.HUMAN:
-            var roomRoster = io.sockets.adapter.rooms[room];
-            if(roomRoster != undefined && roomRoster.length == 1) {
-                return next();
-            } else {
+            var human = client;
+            var bot = botLocator.locateFirstAvailableBotIn(roomName);
+            if(bot == undefined) {
```
takecare: `if (!bot)`?

TelepresenceBot/server-unit/core/server.js
```diff
@@ -72,25 +71,29 @@ io.sockets.on('connection', function (client) {
...
+});
 
-    function leaveAllRooms(client) {
-        var rooms = Object.keys(io.sockets.adapter.sids[client.id]);
-        for(var i = 0; i < rooms.length; i++) {
-            client.leave(rooms[i]);
+function disconnectRoom(name) {
+    var room = io.sockets.adapter.rooms[name];
+
+    if(room != undefined) {
```
takecare: same as  above

TelepresenceBot/server-unit/core/botLocator.js
Mecharyry: I was wondering if that would be better 😆

TelepresenceBot/server-unit/core/botLocator.js
Mecharyry: I always forget that I can do this.

TelepresenceBot/server-unit/core/server.js
Mecharyry: Forgot about this too.

Mecharyry:
Thanks for the review @takecare, I've addressed all of your comments 😄

takecare:
good one @Mecharyry, as always!

PR #276, opened 2017-03-23
Adds basic scripts for code coupling in iOS

amlcurran:
## This PR adds

A few scripts I used whilst evaluating clients' iOS codebases centering around code coupling.

## Considerations and implementation

Three scripts are added:
- `uikit` measures the number of files which import UIKit. This can be used as a (rough) indication that code is improperly coupled to UI libraries.
- `lines` simply gives the number of lines per file. This can then be used to plot a histogram. The point of this is to quickly visualise large files and begin discussions around making them smaller. See an example histogram [here](https://docs.google.com/spreadsheets/d/1m-wIygVxOLJ2KciEokNgpOHeVgIcrM5FwXkoSB-rAyI/pubchart?oid=1766467159&format=interactive)
- `imports` records the number of import statements per file. This can also be used to plot a histogram, and indicates code which is too tightly coupled to other code. See an example histogram [here](https://docs.google.com/spreadsheets/d/1m-wIygVxOLJ2KciEokNgpOHeVgIcrM5FwXkoSB-rAyI/pubchart?oid=472356679&format=interactive)

There are limitations to these files will would be addressed if we feel these are viable:
- `uikit` and `imports` do not walk the tree of their dependencies. Therefore, if class A imports header B, and header B has an import of UIKit, it won't record that class A also implicitly imports UIKit. 
- there is no mechanism to include and exclude directories. Currently, `Pods/` is ignored but it would be nice to enable more customisations

### Test(s) added 

No, this is bash. Should this go further I would consider migrating it to another platform and adding tests there

### Screenshots

image 1
image 2

If you are a novoda employee, you can see the raw data [here](https://docs.google.com/a/novoda.com/spreadsheets/d/1m-wIygVxOLJ2KciEokNgpOHeVgIcrM5FwXkoSB-rAyI/edit?usp=sharing)

### Paired with 

@ no-one

### DIFF
--- a/code-review-tools/README.md
+++ b/code-review-tools/README.md
@@ -1,0 +1,3 @@
+# Code review tools
+
+This aims to serve as a repository for any tools and artifacts that are useful for reviewing code and analysing its effectiveness. Many are simple bash scripts to be run against a repository
--- a/code-review-tools/README.md
+++ b/code-review-tools/README.md
@@ -1,3 +1,19 @@
 # Code review tools
 
-This aims to serve as a repository for any tools and artifacts that are useful for reviewing code and analysing its effectiveness. Many are simple bash scripts to be run against a repository
+This aims to serve as a repository for any tools and artifacts that are useful for reviewing code and analysing its effectiveness.
+
+## Android
+
+No tools currently available.
+
+## iOS
+
+- `uikit` (bash script) measures the number of files which import UIKit. This can be used as a (rough) indication that code is improperly coupled to UI libraries.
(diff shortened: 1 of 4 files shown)


blundell:
Add a readme to `/code-review-tools` , explain the intention of the project etc

amlcurran:
@blundell added - don't really know what should be in it, if you have any recommendations I'm all ears

code-review-tools/README.md
```diff
@@ -0,0 +1,22 @@
+# Code review tools
+
+This aims to serve as a repository for any tools and artifacts that are useful for reviewing code and analysing its effectiveness.
+
+## Android
+
+No tools currently available.
```
takecare: :(

PR #277, opened 2017-03-27
Tpbot/ Web Client

Mecharyry:
#### Problem
The plan for the `MVP` is to open a hangout to connect to the `bot` and chat with people from another office. Well if we are opening a hangout perhaps we should also create a `webClient` so that the user does not need to have both a mobile device for moving the `bot` and another device for chatting?

#### Solution
Create a `webpage` from which the user can navigate the `bot`. This is just the beginning of the `webClient` creating a dummy controller that hands back to our `server` but does not pass any comments through `socketIO` yet. In a follow-up PR I will add hook up `socketIO` and fully test the `webClient` part. 

#### Screen Capture
tpbot_html mov

### DIFF
--- a/TelepresenceBot/server-unit/core/public/js/connect.js
+++ b/TelepresenceBot/server-unit/core/public/js/connect.js
@@ -1,4 +1,19 @@
+var intervalId;
+
+$('body').mousedown(function(event) {
+    console.log("mouseDown");
+    var message = $("input[type=button]").val();
+    intervalId = setInterval(function(){
+        sendMessage(message);
+    }, 500);
+});
+
 function sendMessage(message) {
     var socket = io();
     socket.emit('chat message', message);
 }
+
+$('body').mouseup(function(event) {
+    console.log("mouseUp");
+    clearInterval(intervalId);
+});
--- a/TelepresenceBot/server-unit/core/public/js/connect.js
+++ b/TelepresenceBot/server-unit/core/public/js/connect.js
@@ -7,7 +7,7 @@ $(document).ready(function(){
         var message = $(this).val();
         intervalId = setInterval(function() {
             sendMessage(message);
-        }, 500);
+        }, 100);
     });
 
     function sendMessage(message) {
--- a/TelepresenceBot/server-unit/core/public/js/connect.js
+++ b/TelepresenceBot/server-unit/core/public/js/connect.js
@@ -1,10 +1,5 @@
 var intervalId;
 
-function sendMessage(message) {
-    var socket = io();
-    socket.emit('chat message', message);
-}
-
 $(document).ready(function(){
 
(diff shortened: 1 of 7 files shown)


TelepresenceBot/server-unit/core/testServer.js
```diff
@@ -0,0 +1,45 @@
...
+    console.log("Express server listening on port " + 4200);
+});
+
+app.use(express.static(path.join(__dirname, 'public')));
+
+app.get('/', function(req, res) {
+    res.sendFile(__dirname + '/html/index.html');
+});
+
+app.get('/rooms', function(req, res) {
```
Mecharyry: I wanted to expose a `/rooms` endpoint because I don't want the app to open a `socket` until the app is ready to begin a conversation. Not sure if this is overkill though. I'd like to expose this sort of thing call using `etags` if possible.

TelepresenceBot/server-unit/test/test.js
```diff
@@ -0,0 +1,34 @@
...
+};
+
+describe("echo", function () {
+
+    beforeEach(function (done) {
+        server = require('../core/testServer').server;
+        done();
+    });
+
+    it("echos message", function (done) {
```
Mecharyry: Tests are not complete atm, I did manual testing just to get this off of the ground. Half the problem with this stuff is I need to figure out how to test it 😆  As per the description, I plan on doing the connecting with `socketIO` and testing in the next PR.

TelepresenceBot/server-unit/core/public/js/connect.js
```diff
@@ -0,0 +1,24 @@
+var intervalId;
+
+$(document).ready(function(){
+
+    $("input").mousedown(function(){
+        console.log("mouseDown");
+        var message = $(this).val();
+        intervalId = setInterval(function() {
+            sendMessage(message);
+        }, 500);
```
danybony: ❗️ We should match the `COMMAND_REPEAT_DELAY` defined both in the Android and Arduino parts.
In that case it's 100ms

TelepresenceBot/server-unit/core/testServer.js
danybony: It makes sense, but it depends on the effort required. I think it can be done after the MVP

Mecharyry: For sure. That's why I have added this here and not modified the android app yet 😄

TelepresenceBot/server-unit/core/public/js/connect.js
Mecharyry: Ah yes, I was meant to update this once I got it working 😆

danybony:
💯  @Mecharyry

PR #279, opened 2017-03-29
Tpbot/ Naming and Package structure

Mecharyry:
This PR performs some general cleaning of the codebase in regards to naming and package structure.

- Some classes in `mobile` that are not dependent on `android` and could be moved to core. 
- `com.novoda.tpbot.support` has some classes that are not support classes because they are too tied to the `tpbot` application. Let's move these to the generic `com.novoda.tpbot` package and move both `core` and `mobile` `support` packages to `com.novoda`.
- Rename `AccessibilityService` to better reflect what it actually does which is `HangoutJoinerAutomationService`.
- Rename all instances of `tp` or `tele` etc to `Telepresence` to avoid unnecessary abbreviations.


TelepresenceBot/mobile/src/main/AndroidManifest.xml
```diff
@@ -8,15 +8,15 @@
...
+    android:name="com.novoda.tpbot.TelepresenceBotApplication"
     android:allowBackup="true"
     android:icon="@mipmap/ic_launcher"
     android:label="@string/app_name"
     android:supportsRtl="true"
     android:theme="@style/AppTheme"
     tools:ignore="AllowBackup,GoogleAppIndexingWarning">
 
-    <service android:name="com.novoda.tpbot.automation.TelepresenceBotAccessibilityService"
+    <service android:name="com.novoda.tpbot.automation.HangoutJoinerAutomationService"
```
danybony: 👍

PR #280, opened 2017-03-29
Tpbot/ Open AccessibilitySettings

Mecharyry:
#### Problem
When opening the app the `AccessibilityService` may or may not be enabled and there is no indication that it should be enabled for the app to work correctly.

#### Solution
As a first step, when opening the `BotActivity` detect whether the service is enabled, if not then navigate to the `AccessibilitySettings` screen automatically.

#### Tests added
I've fully tested the `AutomationChecker`, except where it is too close to android.

#### Screen Capture
navigate_to_settings mp4


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/automation/AndroidAccessibilitySettingsRetriever.java
```diff
@@ -0,0 +1,42 @@
...
+        }
+        return EMPTY_STRING;
+    }
+
+    private boolean canRetrieveAccessibilitySettings() {
+        try {
+            return Settings.Secure.getInt(
+                    contentResolver,
+                    Settings.Secure.ACCESSIBILITY_ENABLED
+            ) == 1;
```
ataulm: I don't follow how this try matches the exception thrown in catch. In my understanding the `try` is to check whether _any_ accessibility service is enabled. Could do this with `accessibilityManager.isEnabled()`

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/automation/AndroidAccessibilitySettingsRetriever.java
```diff
@@ -0,0 +1,42 @@
...
+        } catch (Settings.SettingNotFoundException e) {
+            Log.e(AutomationChecker.class.getSimpleName(), "Could not find settings.", e);
```
ataulm: the class used in the Log is not this class - is that intentional?

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/automation/AutomationChecker.java
```diff
@@ -0,0 +1,35 @@
...
+
+        for (String service : accessibilityServices) {
+            if (matchesAutomationService(service)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private boolean matchesAutomationService(String serviceToCheck) {
```
ataulm: this class is generic enough to check whether any service (provided you know the name) is enabled or not. Suggest renaming the class or not passing the serviceName in.

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/automation/AndroidAccessibilitySettingsRetriever.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private final ContentResolver contentResolver;
+
+    public AndroidAccessibilitySettingsRetriever(ContentResolver contentResolver) {
+        this.contentResolver = contentResolver;
+    }
+
+    public String retrieveEnabledAccessibilityServices() {
+        if (canRetrieveAccessibilitySettings()) {
+            return getEnabledAccessibilityServices();
```
ataulm: https://developer.android.com/reference/android/view/accessibility/AccessibilityManager.html#getEnabledAccessibilityServiceList(int) :grin:

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/automation/AndroidAccessibilitySettingsRetriever.java
Mecharyry: Ah lol I did the log and then extracted this logic 😢

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/automation/AndroidAccessibilitySettingsRetriever.java
Mecharyry: So first I decided to remove the feedback flag from the `config.xml` and then I looked at this 😆  For this you **need** to specify at least one feedback mask. So I decided to say that we deal in `Visual` feedback, which is technically true as we perform a click when navigating to the hangouts app. 

This approach made everything a lot simpler 😄  three classes down to one!

ataulm: Do you have to specify a feedback type for your service, even if you pass `FEEDBACK_ALL_MASK` here?

Mecharyry: Yea annoyingly.

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/automation/AutomationChecker.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public AutomationChecker(AccessibilityManager accessibilityManager) {
+        this.accessibilityManager = accessibilityManager;
+    }
+
+    public boolean isHangoutJoinerAutomationServiceEnabled() {
+        List<AccessibilityServiceInfo> visualFeedbackServices = getVisualFeedbackServices();
+
+        for (AccessibilityServiceInfo service : visualFeedbackServices) {
+            if (service.getId().equalsIgnoreCase(SERVICE_ID)) {
```
ataulm: flip the equals? `SERVICE_ID.equals(service.getId())`?

why ignoreCase?

Mecharyry: Just because 😄  done in a004818

PR #281, opened 2017-03-30
[piano-hero] Add data classes to represent notes and chords

ataulm:
We need terminology that can be shared between the sensors (e.g. piano) and actuators (e.g. display).

Since we're not concerned with note timings (yet?), I've only added pitch information. Not sure if we technically _need_ `Chord` since it's just a `Set<Note>` (with an additional constraint of minimum notes == 2) but it's been useful in representing chords that are used to create songs in the `SongSequenceFactory`.

The `SongSequenceFactory` contains examples of songs with single notes and chords too.

### DIFF
--- a/piano-hero/core/src/main/java/com/novoda/pianohero/Chords.java
+++ b/piano-hero/core/src/main/java/com/novoda/pianohero/Chords.java
@@ -1,0 +1,12 @@
+package com.novoda.pianohero;
+
+class Chords {
+
+    static final Chord C = new Chord(Notes.C4, Notes.E4, Notes.G4);
+    static final Chord F_2ND_INV = new Chord(Notes.C4, Notes.F4, Notes.A4);
+    static final Chord G_2ND_INV = new Chord(Notes.D4, Notes.G4, Notes.B4);
+    static final Chord EM = new Chord(Notes.E4, Notes.G4, Notes.B5);
+    static final Chord AM_2ND_INV = new Chord(Notes.E4, Notes.A4, Notes.C5);
+    static final Chord G = new Chord(Notes.G4, Notes.B5, Notes.D5);
+    static final Chord C_2ND_INV = new Chord(Notes.G4, Notes.C5, Notes.E5);
+}
--- a/piano-hero/core/src/main/java/com/novoda/pianohero/Note.java
+++ b/piano-hero/core/src/main/java/com/novoda/pianohero/Note.java
@@ -2,6 +2,9 @@ package com.novoda.pianohero;
 
 public class Note {
 
+    private static final int LOWEST_MIDI_NOTE = 0;
+    private static final int HIGHEST_MIDI_NOTE = 127;
+
     private final int midiNoteNumber;
 
     public Note(int midiNoteNumber) {
@@ -9,8 +12,8 @@ public class Note {
     }
 
     private int checkNotOutOfRange(int midiNoteNumber) {
-        if (midiNoteNumber < 0 && midiNoteNumber > 127) {
-            throw new IllegalArgumentException("MIDI notes range from 0-127 inclusive");
+        if (midiNoteNumber < LOWEST_MIDI_NOTE && midiNoteNumber > HIGHEST_MIDI_NOTE) {
+            throw new IllegalArgumentException("MIDI notes range from " + LOWEST_MIDI_NOTE + "-" + HIGHEST_MIDI_NOTE + " inclusive");
         }
         return midiNoteNumber;
     }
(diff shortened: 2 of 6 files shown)


piano-hero/core/src/main/java/com/novoda/pianohero/Chords.java
```diff
@@ -0,0 +1,12 @@
+package com.novoda.pianohero;
+
+class Chords {
+
+    static final Chord C = new Chord(Notes.C4, Notes.E4, Notes.G4);
+    static final Chord F_2ND_INV = new Chord(Notes.C4, Notes.F4, Notes.A4);
```
ataulm: I dunno the names of these variations off by heart, I checked them [here](http://www.pianochorddictionary.com/chords/g/major-inversion2.shtml)

piano-hero/core/src/main/java/com/novoda/pianohero/Note.java
```diff
@@ -0,0 +1,39 @@
+package com.novoda.pianohero;
+
+public class Note {
+
+    private static final int LOWEST_MIDI_NOTE = 0;
+    private static final int HIGHEST_MIDI_NOTE = 127;
+
+    private final int midiNoteNumber;
+
+    public Note(int midiNoteNumber) {
```
ataulm: MIDI seems to be a really convenient way of representing the pitch since any of the notes on a piano keyboard can be assigned an integer value.

It doesn't make an assumption as to the size of the piano (or even that you're using a piano) which is cool, and we can even derive the associated pitch in hertz for a given note with the midiNoteNumber! :boom:

image

https://en.wikipedia.org/wiki/Scientific_pitch_notation#Table_of_note_frequencies

piano-hero/core/src/main/java/com/novoda/pianohero/Note.java
```diff
@@ -0,0 +1,39 @@
...
+        this.midiNoteNumber = checkNotOutOfRange(midiNoteNumber);
+    }
+
+    private int checkNotOutOfRange(int midiNoteNumber) {
```
ataulm: this is the full range for MIDI - since our app will be limited in scope (C4-B5) should this take into account that range?

Or should we just avoid adding Songs that use notes outside this range? The app would be flexible enough to handle anything in this range.

The limiting factors would be the inputs/Views for output. E.g. you couldn't represent C0 on a treble clef without a ridiculous number of lines and you can't even input C0 using even a full size piano (lowest note is A0).

piano-hero/core/src/main/java/com/novoda/pianohero/Notes.java
```diff
@@ -0,0 +1,29 @@
+package com.novoda.pianohero;
+
+class Notes {
+
+    static final Note C4 = new Note(60);
+    static final Note C4_S = new Note(61);
```
ataulm: S for sharp - `C4_S` == `D4_b` but I didn't see the point in writing `C4_S_D4_b`

piano-hero/core/src/main/java/com/novoda/pianohero/Sequence.java
```diff
@@ -0,0 +1,42 @@
+package com.novoda.pianohero;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+public final class Sequence {
+
+    private final List<Set<Note>> notes;
```
ataulm: I couldn't think of a name for `Set<Note>`. It's the column of notes you'd play together, potentially with both hands. I guess it's still technically a chord? But it could also be just a set of 1, so it could also not be a chord.

piano-hero/core/src/main/java/com/novoda/pianohero/Sequence.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private Sequence(List<Set<Note>> notes) {
```
ataulm: this is implicitly ordered - you have to use the Builder to create this and you don't get access to the list.

ataulm:
nah I've been on different ways to represent this [for a while](https://github.com/ataulm/notes/blob/base_notes/core/src/main/java/com/ataulm/notes/Note.java).

we can split the lib from this - and yours could depend on it? This one uses the discrete set of available midi note numbers, rather than infinite set of frequencies. The `Notes` and `Chords` classes are controversial with naming I think and probably wouldn't want them in a lib.

blundell:
i think its the wrong approach to write things a layer at a time, we should write a thin end to end solution

ataulm:
>i think its the wrong approach to write things a layer at a time, we should write a thin end to end solution

Can do - this PR doesn't affect that though right? For an end-to-end, including user input to match the view output, we need these data classes. Did I misunderstand?

blundell:
its hard to see what this PR achieves, i.e. yes we need data classes but without doing an end to end slice we can't see if these dataclasses are the format we need.

We can use this code for our slice, but i wouldn't merge this PR

ataulm:
ah right, sure - I just wanted to keep PRs small

PR #284, opened 2017-04-03
[piano-hero] Add end-to-end examples

ataulm:
This is a continuation of the [previous PR](https://github.com/novoda/spikes/pull/281) which just added data classes.

## This PR will

add two examples showing end-to-end integration with one output view to display notes (a TextView) and two different inputs (grabbing key input from Activity and a piano view).

It provides a first pass at creation of a "Brain":

- start a `Brain` with a sequence of notes that the user needs to play
- user input is fed _into_ the `Brain`
- the brain will trigger callbacks with the latest Sequence state. The callback (in these cases, our activities) will tell the output view to display the latest Sequence

The Sequence contains the entire list of notes that makes a "song". It also keeps information like "position" (the next set of Notes that the user should play) and "user error" (if the user last input was a mistake, the output view can display this if it likes).

## Screencaps

Keyboard input | Piano input
---|---
simple | piano

## Tests
BrainTest adds tests for the Brain:

- when started with a Sequence, it should callback saying the initial Sequence
- when receiving user input, it should callback
    - if the input is incorrect, the position should stay the same, and the error should be not-empty
    - if the input is correct
        - if that was the last set of notes, callback game complete
        - callback saying next Sequence, with position incremented, empty error


piano-hero/core/src/main/java/com/novoda/pianohero/Notes.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Set;
+
+public class Notes implements Iterable<Note> {
+
+    public static final Notes EMPTY = new Notes(Collections.<Note>emptySet());
+
+    private final Set<Note> notes;
```
ataulm: The utility of the set is to ensure no duplicates - but apart from that, it's a bit annoying because it has no `get(int)` methods...

piano-hero/core/src/main/java/com/novoda/pianohero/Sequence.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public Notes get(int position) {
+        return notes().get(position);
+    }
+
+    public int position() {
+        return position;
+    }
+
+    public Notes latestError() {
```
ataulm: this could totally be an Optional 😂

piano-hero/mobile/src/main/java/com/novoda/pianohero/PianoC4ToB5Activity.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    protected void onPause() {
+        brain.removeCallbacks();
+        inputView.detachKeyListener();
+        super.onPause();
+    }
+
+    private static class NotesPlayedDispatcher implements KeyListener {
```
ataulm: this class will probably be useful for the MIDI controller too, as-is.

piano-hero/mobile/src/main/java/com/novoda/pianohero/PianoC4ToB5View.java
```diff
@@ -0,0 +1,216 @@
...
+        bindKey(g5sView, Note.G5_S);
+        bindKey(a5sView, Note.A5_S);
+    }
+
+    public void detachKeyListener() {
+        this.keyListener = KeyListener.LOGGING;
+    }
+
+    private void bindKey(final View keyView, final Note note, final View... adjacentBlackKeys) {
+        // TODO: Allow user to drag across view to play notes - https://developer.android.com/training/gestures/viewgroup.html
```
ataulm: the dream @mr-archano , less gradle, more views pls.

piano-hero/mobile/src/main/java/com/novoda/pianohero/PianoHeroApplication.java
```diff
@@ -0,0 +1,25 @@
...
+
+    private static Application application;
+
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        application = this;
+    }
+
+    public static void popToast(String message) {
```
ataulm: you all love it.

piano-hero/core/src/main/java/com/novoda/pianohero/Brain.java
```diff
@@ -0,0 +1,73 @@
...
+
+    Brain(OnSequenceUpdatedCallback onSequenceUpdatedCallback) {
+        this.onSequenceUpdatedCallback = onSequenceUpdatedCallback;
+    }
+
+    public void attach(Callback callback) {
+        this.callback = callback;
+    }
+
+    public void removeCallbacks() {
```
jackSzm: Shouldn't it be just `removeCallback` ?

piano-hero/core/src/main/java/com/novoda/pianohero/Brain.java
```diff
@@ -0,0 +1,73 @@
...
+        this.callback = NOOP_CALLBACK;
+    }
+
+    public void start(Sequence sequence) {
+        this.sequence = sequence;
+        onSequenceUpdatedCallback.onNext(sequence);
+    }
+
+    public void onNotesPlayed(Note... notes) {
```
jackSzm: Can you just pass `Notes` ?

ataulm: in most cases we will. the soft keyboard input only allows for one note at a time, so this was easier + for tests.

this is just sugar around the method below which passes notes

piano-hero/mobile/src/main/java/com/novoda/pianohero/PianoC4ToB5View.java
mr-archano: LoL, click bait comment, I see what you did there.

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
```diff
@@ -0,0 +1,105 @@
...
+
+    GameModel(
+            SongSequenceFactory songSequenceFactory,
+            SimplePitchNotationFormatter pitchNotationFormatter) {
+        this.songSequenceFactory = songSequenceFactory;
+        this.pitchNotationFormatter = pitchNotationFormatter;
+    }
+
+    @Override
+    public RoundViewModel startGame() {
```
blundell: ya so we should make this return `void` and Callback the ViewModel

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
```diff
@@ -0,0 +1,105 @@
...
+                return "";
+            }
+        } else {
+            return "Ruhroh, try again!";
+        }
+    }
+
+    @Nullable
+    @Override
+    public RoundViewModel onNotesPlayed(Callback callback, Note... notes) {
```
blundell: when this returns `void` we can have two separate callbacks. Therefore two separate methods. Perhaps:

`hasGameFinished(Callback`
and
`playGameRound(Notes, Callback`

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
```diff
@@ -0,0 +1,36 @@
+package com.novoda.pianohero;
+
+interface GameMvp {
+
+    interface Model {
+
+        void startGame(StartCallback callback);
+
+        RoundViewModel onNotesPlayed(CompletionCallback callback, Note... note);
```
ataulm: should the model know the viewmodel? In my understanding of MVP, M and V do not meet.

It would be the presenter or a layer between the presenter and the View that creates the ViewModel.

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
```diff
@@ -0,0 +1,36 @@
...
+        void showRound(RoundViewModel viewModel);
+
+        void showGameComplete();
+    }
+
+    interface Presenter {
+
+        void onCreate();
+
+        void onNotesPlayed(Note... notes);
```
ataulm: similarly (though it could be argued we're shortcutting), neither the view nor model should know about the presenter. In this example, the View knows the presenter.

should these methods just be **separate** callbacks that the view knows about (/owns), _provided_ by the presenter?

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
```diff
@@ -0,0 +1,101 @@
...
+            for (Note note : notes.notes()) {
+                if (pitchNotationFormatter.format(note).endsWith("#")) {
+                    throw new IllegalArgumentException("Sequence contains sharps, that's not simple enough");
+                }
+            }
+        }
+    }
+
+    @NonNull
+    private RoundViewModel createViewModel(Sequence updatedSequence) {
```
ataulm: should the model know the viewmodel? In my understanding of MVP, M and V do not meet/know _anything_ about each other. In this interface, the model knows about the view, and so it breaks separation of concerns that we tried to achieve?

It would be the presenter or a layer between the presenter and the View that creates the ViewModel.

---

Otherwise, RoundViewModel is actually a domain model that's shared, and it shouldn't be called ViewModel.

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
blundell: > The Presenter itself shouldn't be known to the View, since it's the "master" of the View/Model.

I'm not sure it matters if it knows it or not. Whats the benefit of "hiding" the presenter?

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
blundell: yeah there would be a Converter that converts in the model from the Domain objects to the ViewModel, hasn't been done yet. I've still been trying to understand the Domain stuff you wrote :-).

Like this: https://github.com/novoda/yourmd-android/blob/master/learn/src/main/java/md/your/learn/v3/ArticleGroupsModel.java#L36

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
ataulm: In my mind, it'll be reducing the scope of the View's knowledge, keeping it dumber.

It's just naming anyway, since here the Presenter is an interface and it'd contain all the methods that I'm suggesting should be in the View, it won't make a big difference; just conceptually.

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
ataulm: The converter is fine, though if the converter is inside the model, it still means the model knows about the view.

model | presenter | view

I mean that the viewmodel should only exist to the right of the model in the diagram I spent ages writing above ^^^

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
blundell: i think adding a callback might make it more complex, can we leave it for now?

blundell:
:shipit:

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
blundell: technically the `Model` won't know about it, the `Callback<ViewModel>` and the `Converter<Domain, ViewModel>` will

ataulm: The Model knows about it if the conversion happens in the Model class

blundell: the Types are erased at run time :trollface: 

`converter.convert(domainModel)` has to happen somewhere, just have to deal with it that its here (as this will be off the mainthread SOON) any higher up will be MT

ataulm: Why would it matter whether it's main thread or not?

ataulm: also the types aren't erased at runtime, that would be for generics right?

blundell: yeah the generic types `Callback<ViewModel>` 


 The point about MT was saying you want to do as much as possible in the background, for dat smooth UI

PR #285, opened 2017-04-03
move to an mvp format

blundell:
Create a Model , View Presenter format.

There is still a lot we can improve with our separations.

We could reintroduce the "Brain" although I would call it "GameEngine" and this would be used by the model, but what we have here is the minimum for MVP

it makes the VIew super dumb and just display data
it makes the Activity do nothing except convert from Android (controversial)
it has the presenter co-ordinate between View & Model
And the Model does all of the game logic

At one point we return null, which sucks, I left a TODO

There is no concept of converters yet (to convert from domain model to view model) and the `Sequence` is used in both which is probably wrong.


Mecharyry:
Wow that description :trollface:

blundell:
@Mecharyry heh sorry about that, updated!

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
```diff
@@ -0,0 +1,104 @@
...
+
+public class GameModel implements GameMvp.Model {
+
+    private final SongSequenceFactory songSequenceFactory;
+    private final SimplePitchNotationFormatter pitchNotationFormatter;
+
+    private Sequence sequence;
+
+    GameModel(
+            SongSequenceFactory songSequenceFactory,
```
ataulm: I can live with it all but this indentation? my goodness

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
```diff
@@ -0,0 +1,104 @@
...
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
+public class GameModel implements GameMvp.Model {
+
+    private final SongSequenceFactory songSequenceFactory;
```
ataulm: :+1:

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleEndToEndActivity.java
```diff
@@ -60,35 +34,24 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
-
     @Override
     public boolean onKeyUp(int keyCode, KeyEvent event) {
         try {
             Note note = keyCodeConverter.convert(keyCode);
-            brain.onNotesPlayed(note);
+            presenter.onNotesPlayed(note);
         } catch (IllegalArgumentException e) {
-            PianoHeroApplication.popToast("that's not a simple note!");
+            Log.e("!!!", "Non supported key input");
```
ataulm: no user feedback if they press the wrong keys

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleNotesOutputView.java
```diff
@@ -29,54 +26,32 @@ protected void onFinishInflate() {
...
     }
 
-    private void updateNextNoteToPlay(Sequence sequence) {
-        int nextNotesPosition = sequence.position();
-        for (Note note : sequence.get(nextNotesPosition)) {
-            updateNextNoteToPlayTextViewWith(note);
-        }
+    @Override
+    public void showGameComplete() {
+        PianoHeroApplication.popToast("game complete, another!");
```
ataulm: this message should be passed as a parameter - the presenter is the only one that knows that you're starting a new game onGameComplete

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleNotesOutputView.java
```diff
@@ -29,54 +26,32 @@ protected void onFinishInflate() {
...
     }
 
-    private void updateNextNoteToPlayTextViewWith(Note note) {
-        String simpleNotation = simplePitchNotationFormatter.format(note);
-        nextNoteToPlayTextView.setText(simpleNotation);
+    @Override
+    public void showError() {
+        PianoHeroApplication.popToast("that's not a simple note!");
```
ataulm: :raised_hands:

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleNotesOutputView.java
```diff
@@ -29,54 +26,32 @@ protected void onFinishInflate() {
...
-                String message = String.format(Locale.US, "Woo! Keep going! (%d/%d)", sequence.position() + 1, sequence.length());
-                updateMessageTextViewWith(message);
-            } else {
-                messageTextView.setVisibility(INVISIBLE);
-            }
-        } else {
-            updateMessageTextViewWith("Ruhroh, try again!");
-        }
+    @Override
+    public void showKeyboard() {
```
ataulm: GameMvp.View has a method called `showKeyboard` which is dependent on platform/view. Most of the Views will not support this

piano-hero/mobile/src/main/java/com/novoda/pianohero/PianoC4ToB5Activity.java
```diff
@@ -1,64 +1,34 @@
...
-        @Override
-        public void onNext(Sequence sequence) {
-            outputView.display(sequence);
-        }
-    };
-
-    private void startNewGame() {
-        Sequence sequence = new SongSequenceFactory().maryHadALittleLamb();
-        brain.start(sequence);
+        GameMvp.View outputView = (SimpleNotesOutputView) findViewById(R.id.simple_notes_output_view);
```
ataulm: :x: this is the wrong view for this activity - it's supposed to be the PianoC4ToB5View

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
blundell: @eduardb he's mad

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleEndToEndActivity.java
blundell: you cross it out, but just to confirm - no user feedback because we never expect to allow non supported input

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
```diff
@@ -0,0 +1,104 @@
...
+            } else {
+                return "";
+            }
+        } else {
+            return "Ruhroh, try again!";
+        }
+    }
+
+    @Override
+    public RoundViewModel onNotesPlayed(Callback callback, Note... notes) {
```
ataulm: :x: this makes it synchronous - the callback was being used to pump out events:

- allows it to be asynchronous
- means we don't return null (also this method isn't marked with `Nullable`) because we return nothing

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
```diff
@@ -0,0 +1,38 @@
...
+        }
+    }
+
+    interface View {
+
+        void showRound(RoundViewModel viewModel);
+
+        void showGameComplete();
+
+        void showError();
```
ataulm: :x: we don't pass the error anymore - we lose the ability to show the user what the error was

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleNotesOutputView.java
blundell: yes if it got more complicated (or wanted to test) you could pass a `GameCompleteViewModel`

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleNotesOutputView.java
blundell: yes so maybe a better name is `showInput` I was thinking keyboard as in the domain wording 

probably the real answer is get rid of the "show keyboard" menu item

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
blundell: the asynchronisity has moved to the presenter

piano-hero/mobile/src/main/java/com/novoda/pianohero/PianoC4ToB5Activity.java
blundell: hm i didn't change this: https://github.com/novoda/spikes/pull/284/files#diff-6d78cbc1ad465d96123adaa06b0aced7R26

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
blundell: I wouldn't X this, what you are saying is a next step 👍

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
blundell: @Electryc but should it? :D

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
ataulm: it's a next step that already exists, it's regression

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
blundell: added annotation https://github.com/novoda/spikes/pull/285/commits/e57dd1804a91318cace5fe2dd90147a9eb71af9d

piano-hero/mobile/src/main/java/com/novoda/pianohero/PianoC4ToB5Activity.java
ataulm: yes you're right :+1:

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleEndToEndActivity.java
ataulm: I crossed it out because the feedback (the toast) still occurs elsewhere.

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleNotesOutputView.java
ataulm: need to have it because you can close the keyboard and not open it again.

ataulm: `showInput` is implementation specific too - in the case of AndroidThings buttons or MIDI controller it'll be no-op?

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleNotesOutputView.java
```diff
@@ -4,14 +4,11 @@
...
+import android.view.inputmethod.InputMethodManager;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
-import java.util.Locale;
-
-public class SimpleNotesOutputView extends LinearLayout {
-
-    private final SimplePitchNotationFormatter simplePitchNotationFormatter = new SimplePitchNotationFormatter();
+public class SimpleNotesOutputView extends LinearLayout implements GameMvp.View {
```
ataulm: a separate java class that implements GameMvp.View might work. That class could take this AndroidView as a collaborator, but then could perform the other things on its own. E.g. the showKeyboard/Input, doesn't need to be part of this AndroidView.

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
```diff
@@ -0,0 +1,105 @@
...
+    public RoundViewModel startGame() {
+        sequence = songSequenceFactory.maryHadALittleLamb();
+
+        checkSequenceIsSimpleElseThrow(sequence);
+
+        return createViewModel(sequence);
+    }
+
+    private void checkSequenceIsSimpleElseThrow(Sequence sequence) {
+        for (int i = 0; i < sequence.length(); i++) {
```
Electryc: Is `Sequence` `Iterable`?  If so we could use the enhanced for loop 😄

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public RoundViewModel onNotesPlayed(Callback callback, Note... notes) {
+        return onNotesPlayed(new Notes(notes), callback);
+    }
+
+    @Nullable
+    private RoundViewModel onNotesPlayed(Notes notes, Callback callback) {
+        if (notes.count() == 0) {
+            callback.onSequenceComplete();
+            return null;
```
Electryc: What does a `null` RoundViewModel mean? Can't we have a representation of it in the view model? Like an empty version

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
```diff
@@ -0,0 +1,38 @@
...
+
+        void showKeyboard();
```
Electryc: Who do you expect to implement this `View`, I see it platform agnostic. So `showKeyboard()` here is not really platform agnostic. Maybe we're missing something. What's the domain action of this command if we're not thinking platform specific?

Also, being in `mobile` it doesn't force you to think about it as platform agnostic.

piano-hero/mobile/src/main/java/com/novoda/pianohero/RoundViewModel.java
```diff
@@ -0,0 +1,23 @@
+package com.novoda.pianohero;
+
+import java.util.Iterator;
+import java.util.List;
+
+class RoundViewModel implements Iterable<String> {
```
Electryc: Aha! 😄

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleEndToEndActivity.java
```diff
@@ -1,54 +1,28 @@
...
 import android.os.Bundle;
 import android.support.annotation.Nullable;
 import android.support.v7.app.AppCompatActivity;
+import android.util.Log;
 import android.view.KeyEvent;
 import android.view.Menu;
 import android.view.MenuItem;
-import android.view.inputmethod.InputMethodManager;
 
 public class SimpleEndToEndActivity extends AppCompatActivity {
```
Electryc: This activity name 😂

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
```diff
@@ -0,0 +1,105 @@
...
+
+        int currentPosition = sequence.position();
+        Notes expectedNotes = sequence.get(currentPosition);
+        if (currentPosition == sequence.length() - 1 && notes.equals(expectedNotes)) {
+            callback.onSequenceComplete();
+            return null; // TODO we should separate the querying of complete, so that null return is no necessary
+        }
+
+        if (notes.equals(expectedNotes)) {
+            this.sequence = new Sequence.Builder(sequence).atPosition(currentPosition + 1).build();
```
Electryc: Is this `sequence` reflecting the current state of the game ? As in the notes that have been played up until now.

piano-hero/mobile/src/main/java/com/novoda/pianohero/GamePresenter.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public void onCreate() {
+        RoundViewModel viewModel = gameModel.startGame();
+        view.showRound(viewModel);
+    }
+
+    @Override
+    public void onNotesPlayed(Note... notes) {
+        RoundViewModel viewModel = gameModel.onNotesPlayed(new GameMvp.Model.Callback() {
```
Electryc: So the whole business logic happens in the `model`, and it callbacks to the presenter when the game succeeds so that the presenter knows what to do going forward 👍 

What I find it weird is that this method `onNotesPlayed` returns a viewModel synchronously, but at the same time attaches this callback that async says when the game is complete. It seems to me that the attaching of the callback could be done separately ? 

`gameModel.setOnSequenceComplete(new GameMvp.Model.Callback())`

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleNotesOutputView.java
blundell: maybe or : http://stackoverflow.com/questions/1509719/android-how-to-make-the-keypad-always-visible so this method would not be needed 

___ 

Alternatively yeah you can close the keyboard and not open it again, so I think like you said - we should have separate activities and therefore separate Views & Presenters , perhaps all using the same Model

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
blundell: I don't see it happening in the code? perhaps because of https://github.com/novoda/spikes/pull/285#discussion_r109465194

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
blundell: yeah I put a TODO on the next return of null, at the moment the presenter catches the null and the onComplete callback is called. So to the user the game is complete.

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
blundell: yes. not this PR tho :)

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    private void checkSequenceIsSimpleElseThrow(Sequence sequence) {
+        for (int i = 0; i < sequence.length(); i++) {
+            Notes notes = sequence.get(i);
+            if (notes.count() > 1) {
+                throw new IllegalArgumentException("Sequence contains chords, that's not simple enough");
+            }
+            for (Note note : notes.notes()) {
+                if (pitchNotationFormatter.format(note).endsWith("#")) {
```
blundell: i think in the future Ataul will be looking at this, this PR was just a refactor to mvp

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
blundell: thats my understanding as well

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
blundell: what if I told you this was a _keyboard_

piano-hero/mobile/src/main/java/com/novoda/pianohero/GamePresenter.java
blundell: yeah it was a hack to do the refactoring with least damage. It's the same answer as the `return null`

`// TODO we should separate the querying of complete, so that null return is no necessary`

I will look at separating this now in this PR, to avoid confusion

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
eduardb: @ataulm  mañana te mostraré algo 😄

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
Electryc: Ha! Actually after writing this, I was thinking what if _keyboard_ is actually part of the domain regardless the delivery mechanism (which is the _view_ implementation). At the time of writing my previous comment, I was thinking `keyboard` as in Android keyboard with letters and numbers on it

blundell: heh but now Ataul has spoken to me, he's convinced me of your original comment because this is all view layer without any logic in the model, so I've inlined this back into the activity

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
ataulm: The SimplePitchNotationFormatter outputs as if key of C so only sharps

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameModel.java
ataulm: it includes all notes past/present - the `atPosition(int position)` is the pointer to show what's been played so far.

piano-hero/mobile/src/main/java/com/novoda/pianohero/SimpleEndToEndActivity.java
```diff
@@ -34,7 +36,7 @@ public boolean onCreateOptionsMenu(Menu menu) {
     @Override
     public boolean onOptionsItemSelected(MenuItem item) {
         if (item.getItemId() == R.id.show_keyboard) {
-            presenter.onShowKeyboardSelected();
+            showKeyboard();
```
ataulm: 👍 will blow your mind in next PR

piano-hero/mobile/src/main/java/com/novoda/pianohero/GamePresenter.java
blundell: meh, too big a change for this PR - which is just refactoring, will ping you on next 1!

piano-hero/mobile/src/main/java/com/novoda/pianohero/GameMvp.java
ataulm: ah man I came to comment on this and found it's resolved already..

PR #286, opened 2017-04-05
DropCap/ Allow multiple DropCaps

Mecharyry:
#### Problem
As per issue #282 a `DropCap` cannot be applied to more than a single character. The most pressing example where this would be an issue is in the following (pulled from issue), where a normally `DropCapped` character is preceded by a quotation mark.

image

#### Solution
Allow the client to specify the number of characters to `DropCap`. 

In the event that the copy wrapping the `DropCap` cannot fit on the remaining width of the `viewPort` the `DropCap` is not drawn. Typically this will only occur if the client specifies too many `DropCaps` such that the remaining width (to the right of the DropCap) is less than the size of one character of copy text.

#### Screen Captures

Single DropCap | Two DropCaps
---|---
drop_cap_one | drop_cap_two

Lots of DropCaps | Not enough room for copy
---|---
drop_cap_all | drop_cap_too_many

### DIFF
--- a/drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
+++ b/drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
@@ -88,7 +88,7 @@ public class DropCapView extends View {
             setCopyTextColor(copyTextColor);
 
             String text = typedArray.getString(R.styleable.DropCapView_android_text);
-            setText(text);
+            setText(text, numberOfDropCaps);
 
         } finally {
             typedArray.recycle();
@@ -96,8 +96,8 @@ public class DropCapView extends View {
     }
 
     public void setNumberOfDropCaps(int numberOfDropCaps) {
-        this.numberOfDropCaps = numberOfDropCaps;
-        remeasureAndRedraw();
+        String text = dropCapText + copyText;
+        setText(text, numberOfDropCaps);
     }
 
     public void setDropCapFontType(String fontPath) {
@@ -212,7 +212,11 @@ public class DropCapView extends View {
     }
 
     public void setText(String text) {
-        if (isSameText(text)) {
+        setText(text, numberOfDropCaps);
+    }
+
+    private void setText(String text, int numberOfDropCaps) {
+        if (sameNumberOfDropCaps(numberOfDropCaps) && sameText(text)) {
             return;
         }
 
@@ -224,14 +228,19 @@ public class DropCapView extends View {
             copyText = (text == null) ? "" : text;
         }
 
+        this.numberOfDropCaps = numberOfDropCaps;
         remeasureAndRedraw();
     }
 
-    private boolean isSameText(String text) {
+    private boolean sameNumberOfDropCaps(int numberOfDropCaps) {
+        return this.numberOfDropCaps == numberOfDropCaps;
+    }
+
+    private boolean sameText(String text) {
         return text != null && text.equals(dropCapText + copyText);
     }
 
-    private boolean enoughTextForDropCap(CharSequence text) {
+    private boolean enoughTextForDropCap(String text) {
         return text != null && text.length() > numberOfDropCaps;
     }
 
@@ -257,7 +266,7 @@ public class DropCapView extends View {
 
     private void measureDropCapFor(int width) {
         dropCapWidth = (int) (dropCapPaint.measureText(dropCapText, 0, dropCapText.length()) + spacer);
-        dropCapPaint.getTextBounds(dropCapText, 0, numberOfDropCaps, dropCapBounds);
+        dropCapPaint.getTextBounds(dropCapText, 0, dropCapText.length(), dropCapBounds);
         int copyWidthForDropCap = width - dropCapWidth;
 
         if (dropCapStaticLayout == null || dropCapStaticLayout.getWidth() != copyWidthForDropCap) {
@@ -383,7 +392,7 @@ public class DropCapView extends View {
             int baseline = dropCapStaticLayout.getLineBaseline(i) + getPaddingTop();
 
             if (i == 0) {
-                lineStart = lineStart + 1;
+                lineStart = lineStart + dropCapText.length();
             }
 
             canvas.drawText(
--- a/drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
+++ b/drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
@@ -32,7 +32,6 @@ public class DropCapView extends View {
     private String dropCapText;
     private String copyText;
 
-    private int numberOfDropCaps;
     private int lineSpacingExtra;
     private int numberOfLinesToSpan;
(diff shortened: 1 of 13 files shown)


drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
```diff
@@ -202,30 +231,6 @@ public int getCopyTextColor() {
         return copyTextPaint.getColor();
     }
 
-    public void setText(String text) {
-        if (isSameText(text)) {
```
Mecharyry: `isSameText` was dropped as an unnecessary optimisation and makes the code easier to read when considering that the `setNumberOfDropCaps` also needs to call this method.

drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
```diff
@@ -246,49 +251,61 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
...
             );
 
             calculateLinesToSpan();
+            calculateDropCapBaseline();
 
-            if (enoughLinesForDropCap()) {
-                float baseline = dropCapBounds.height() + getPaddingTop();
-                dropCapBaseline = baseline - dropCapBounds.bottom;
-            }
+            canDrawDropCap = enoughLinesToWrapDropCap() && dropCapCopyFitsWithin(totalWidth);
```
Mecharyry: The rules determining whether a `DropCap` can be drawn are only present in the `measureDropCapFor` now instead of relying on checks to `enoughLinesToWrapDropCap` all over the codebase.

drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
```diff
@@ -246,49 +251,61 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
...
         }
     }
 
     private void calculateLinesToSpan() {
         int currentLineTop = 0;
         numberOfLinesToSpan = 0;
 
-        for (int i = 0; i < dropCapStaticLayout.getLineCount(); i++) {
-            currentLineTop = dropCapStaticLayout.getLineTop(i);
+        for (int i = 0; i < dropCapCopyStaticLayout.getLineCount(); i++) {
```
Mecharyry: Rename because `dropCapStaticLayout` implied that it was for the `DropCap` which it is not. It is actually used for the copy wrapping the `DropCap`.

drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
```diff
@@ -246,49 +251,61 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
...
+        dropCapWidth = (int) (dropCapPaint.measureText(dropCapText, 0, dropCapText.length()) + spacer);
+        dropCapPaint.getTextBounds(dropCapText, 0, dropCapText.length(), dropCapBounds);
+        int remainingWidthAfterDropCap = totalWidth - dropCapWidth;
 
-        if (dropCapStaticLayout == null || dropCapStaticLayout.getWidth() != copyWidthForDropCap) {
-            dropCapStaticLayout = new StaticLayout(
-                    dropCapText + copyText,
+        if (dropCapCopyStaticLayout == null || dropCapCopyStaticLayout.getWidth() != remainingWidthAfterDropCap) {
+            dropCapCopyStaticLayout = new StaticLayout(
+                    copyText,
```
Mecharyry: Unnecessarily relying on the `DropCapText` and `copyText` when this only deals with the remaining text, the whole text minus the `DropCap` text. 

This makes the drawing process easier for the copy wrapping the `DropCap` because it does not need to check the start position and offset it by the `DropCapText.length()`.

drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
```diff
@@ -341,58 +358,54 @@ protected void onDraw(Canvas canvas) {
...
         for (int i = 0; i < numberOfLinesToSpan; i++) {
-            int lineStart = dropCapStaticLayout.getLineStart(i);
-            int lineEnd = dropCapStaticLayout.getLineEnd(i);
+            int lineStart = dropCapCopyStaticLayout.getLineStart(i);
+            int lineEnd = dropCapCopyStaticLayout.getLineEnd(i);
 
-            int baseline = dropCapStaticLayout.getLineBaseline(i) + getPaddingTop();
-
-            if (i == 0) {
-                lineStart = lineStart + 1;
```
Mecharyry: This was done to offset the `lineStart` to be after the end of the `DropCap` text but this is easily avoided if the `StaticLayout` only deals with the remaining text.

Mecharyry:
I hadn't thought about adding the `@Nullable` @tasomaniac but now that you mention it 😄  All of the methods taking a `non-primitive` have `null` guards by either throwing or returning a default 👍

drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
```diff
@@ -64,6 +68,9 @@ private void applyCustomAttributes(Context context, AttributeSet attrs) {
             String copyFontPath = typedArray.getString(R.styleable.DropCapView_copyFontPath);
             setCopyFontType(copyFontPath);
 
+            int defaultNumberOfDropCaps = 1;
```
eduardb: I would put this into a constant 😛

drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
```diff
@@ -91,7 +98,30 @@ private void applyCustomAttributes(Context context, AttributeSet attrs) {
...
+        this.numberOfDropCaps = numberOfDropCaps;
+
+        String text = dropCapText + copyText;
+        setText(text);
+    }
+
+    public void setText(@Nullable String text) {
+        if (enoughTextForDropCap(text)) {
+            dropCapText = String.valueOf(text.substring(0, numberOfDropCaps));
+            copyText = String.valueOf(text.subSequence(dropCapText.length(), text.length()));
```
eduardb: Any reason for not using directly `numberOfDropCaps` instead of `dropCapText.length()`?

Mecharyry: I did this more to avoid the pain of experienced when putting the `numberOfDropCaps` in again in the future 😆  when I was adding `numberOfDropCaps` I needed to go through and take out `0s` and `1s` I wanted to avoid that by using the actual text as it is something that will always be present.

drop-cap/library/src/main/java/com/novoda/dropcap/DropCapView.java
Mecharyry: Yea I could 😆  I did this to match the others and wasn't sure I wanted a constant that was only in one place to live for that long.

PR #289, opened 2017-04-09
Tpbot/Bot presentation in service

Mecharyry:
#### Problem
The `Bot` is able to automatically open a `hangout` when connected to the middleman server. When opening the `hangout` the `BotActivity` is torn down including any necessary presenters. Tearing down the presentation layer of the `Bot` automatically disconnects said `Bot` from the middleman server. 

#### Solution
Create a `BoundService` and perform all `Bot` actions in this service. For now, only the logic related to connecting and listening for moves has been moved. In the next PR the movement service will be added to the `BoundService` rather than having two independent services (if possible).
The unbinding of the `BotService` is performed in `onDestroy` ensuring that the service lives even when the app is moved to the background (similar how multi-window is handled). 

#### Screen Capture
No updates.

### DIFF
--- a/TelepresenceBot/mobile/src/main/AndroidManifest.xml
+++ b/TelepresenceBot/mobile/src/main/AndroidManifest.xml
@@ -16,7 +16,8 @@
     android:theme="@style/AppTheme"
     tools:ignore="AllowBackup,GoogleAppIndexingWarning">
 
-    <service android:name="com.novoda.tpbot.automation.HangoutJoinerAutomationService"
+    <service
+      android:name="com.novoda.tpbot.automation.HangoutJoinerAutomationService"
       android:permission="android.permission.BIND_ACCESSIBILITY_SERVICE">
       <intent-filter>
         <action android:name="android.accessibilityservice.AccessibilityService" />
@@ -58,6 +59,10 @@
 
     <service
       android:name="com.novoda.tpbot.bot.MovementService"
+      android:exported="false" />
+
+    <service
+      android:name="com.novoda.tpbot.bot.BotService"
       android:exported="false" />
 
   </application>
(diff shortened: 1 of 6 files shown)


TelepresenceBot/mobile/src/main/AndroidManifest.xml
```diff
@@ -60,6 +61,10 @@
       android:name="com.novoda.tpbot.bot.MovementService"
       android:exported="false" />
 
+    <service
+      android:name="com.novoda.tpbot.bot.BotService"
+      android:exported="false" />
+
```
jackSzm: 👍

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/BotActivity.java
```diff
@@ -174,10 +174,15 @@ protected void onStop() {
             unbindService(serviceConnection);
             boundToMovementService = false;
         }
-        presenter.stopPresenting();
         super.onStop();
     }
 
+    @Override
+    protected void onDestroy() {
```
jackSzm: So basically we would still close the connection when activity is destroyed ?

So having this is mind when device would low run in memory and decide to kill the activities that are in the background , then the service would be destroyed as well right ?

Mecharyry: Yes exactly. We want it to be available as long as the activity is bound in the activity backstack. If the phone runs out of memory / the activity is killed by the user then everything is dropped, including the service.

jackSzm: LGTM then !

PR #292, opened 2017-04-14
[piano-hero] Move interfaces to core

ataulm:
The interfaces were all in `mobile` but these all were business concepts so could live in core.

The only issue was that the KeyListener interface had the `android.Log` import. Since this was only used in one View, I moved the logging implementation there.


blundell:
2 days ago! shoulda whatsapp'd me :)

piano-hero/core/src/main/java/com/novoda/pianohero/KeyListener.java
```diff
@@ -0,0 +1,8 @@
+package com.novoda.pianohero;
+
+interface KeyListener {
```
blundell: i'm not sure if this should be in core. it means _key_ as in _keyboard_ right, what about a voice interface? that would not have a key, therefore shouldn't this stay out of core?

ataulm: rename to NoteListener probably.

ataulm: commit b86cc27fc4eb0c73fd4903956dd801b1f4806b77
Author: Ataul Munim <hello@ataulm.com>
Date:   Tue Apr 18 09:21:08 2017 +0100

    Rename KeyListener to NoteListener

blundell:
it still doesn't need to be in `core` it's part of the Android View. The place where it gets sent to our core is `presenter.onNotesPlayed`

ataulm:
ah yeah you're right. The "NotesDispatcher" should also be in the `View` otherwise the presenter will have to handle different type of input

- user hold c
- presenter gets c
- user hold e
- presenter gets e, treats as {c e}
- user hold g
- presenter gets g, treats as {c e g}

rather than the clean "check this set of Notes":

- user hold c
- presenter gets {c}
- user hold e
- presenter gets {c e}
- user hold g
- presenter gets {c e g}

PR #298, opened 2017-05-22
Tpbot/network tweak

danybony:
Two small tweaks to make the network communication between human and bot working:

* Forward the commands received from the network to the serial
* Reduce the delay between repeated commands (so that the arduino doesn't go in timeout and keep executing the previously sent command)


blundell:
not much to comment on :D

blundell:
2 thumbs up, murrrrge

PR #299, opened 2017-05-22
Tpbot/save last server

danybony:
It's incredibly tedious to write every time the Socket IO address, so we are now storing it to shared prefs once we manage to successfully connect


TelepresenceBot/mobile/src/main/res/layout/switchable_server_declaration.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<com.novoda.tpbot.human.ServerDeclarationView xmlns:android="http://schemas.android.com/apk/res/android"
+<com.novoda.tpbot.controls.ServerDeclarationView xmlns:android="http://schemas.android.com/apk/res/android"
```
Mecharyry: 👍

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/BotActivity.java
```diff
@@ -201,10 +208,11 @@ public void onServiceDisconnected(ComponentName componentName) {
     };
 
     @Override
-    public void onConnect(String room) {
+    public void onConnect(String room, String serverAddress) {
         debugView.showPermanently(getString(R.string.connected));
         switchableView.setDisplayedChild(1);
 
+        lastServerPreferences.saveLastConnectedServer(serverAddress);
```
blundell: is this a temporary hack?

otherwise : I would not be saving state inside the View, I would do it inside of the model (looks like it's called a service in your case). 

i.e. inside of `humanTelepresenceService.connectTo(serverAddress)` compose the `humanTelepresenceService` of SharedPreferences/LongtermStorage

Mecharyry: 💯 This should live in the presenter ideally. We should create an interface, because all of the presenters live in core and preferences is an implementation detail based on the environment, in this case, Android.  The presenter can then be responsible for telling the views etc of the initial server address and passing it off to the data layer. Thanks for mentioning this @blundell I didn't pick up on it, been too long away from this.

danybony: Good points @blundell, @Mecharyry.
I extracted an interface for the server address persistence and moved the two save invocations to the presenters (adding tests).
Have a look at https://github.com/novoda/spikes/pull/299/commits/2db5c21bd9c98ae19a6ddea1f1ff3b8580ef3ac9 and https://github.com/novoda/spikes/pull/299/commits/7d467488933c7ddaf9155a2f4f516a4d91046713

TelepresenceBot/core/src/main/java/com/novoda/tpbot/human/HumanPresenter.java
```diff
@@ -42,6 +47,7 @@ public void update(Result result) {
             if (result.isError()) {
                 humanView.onError(result.exception().get().getMessage());
             } else {
+                lastServerPersistence.saveLastConnectedServer(serverAddress);
```
blundell: this is fine 👍 , but you know a presenter is the "glue" between the model and the view.

you could argue that what you have here is still just glueing the two together, (the model being `HumanTelepresenceService` & `LastServerPersistence`) as it gets more complicated your presenter gets more and more collaborators.

If your presenter had only one dependency something like `HumanModel` (and this was composed of `HumanTelepresenceService` & `LastServerPersistence`) then the presenter stays relatively simple "gluing" the model to the view.
It will be your model that expands and gets bigger and bigger (which is awesome because this is where your domain is.)  Then if you decided "ok now we save the last connected address when the connection is established but I actually want to save it when the connection is attempted" the change would be made in the model (currently the change would be made in the presenter), this is clearly a domain & model decision - not the presenters, who should just be presenting data for the UI to be shown.

Food for thought :-)

Mecharyry: 🤔  interesting, I like the idea. Do you have an example? I would like to try the idea, I can't think of any drawbacks without trying it 👅

blundell: will DM you a private project :)

PR #301, opened 2017-05-27
[piano-hero] Replace KeyListener with NoteListener and plug midi support into Piano...Activity

ataulm:
This PR modifies the `KeyStationMini32` class to use `NoteListener` as a callback instead of `KeyListener` so we can plug it into `PianoC3ToB5Activity` to test if the game works.

It pulls out a `SimpleUsbMidiDriver` because it has so many unused methods too. It seems the driver doesn't work when you use it on a regular Android device to connect a digital piano - perhaps we'd have to check the MIDI feature thing and use the normal android feature where available, and the USB one otherwise.

### DIFF
--- a/piano-hero/mobile/src/main/java/com/novoda/pianohero/AndroidThingsActivity.java
+++ b/piano-hero/mobile/src/main/java/com/novoda/pianohero/AndroidThingsActivity.java
@@ -7,19 +7,24 @@ import android.util.Log;
 
 public class AndroidThingsActivity extends AppCompatActivity {
 
+    private final SimplePitchNotationFormatter simplePitchNotationFormatter = new SimplePitchNotationFormatter();
     private MidiKeyboardDriver midiKeyboardDriver;
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         Log.d("!!!", "I'm running");
-
         midiKeyboardDriver = new MidiKeyboardDriver.KeyStationMini32(this);
-        midiKeyboardDriver.attachListener(new MidiKeyboardDriver.KeyListener() {
+        midiKeyboardDriver.attachListener(new NoteListener() {
 
             @Override
-            public void onKeyPressed(Note note) {
-                Log.d("!!!", "Note pressed " + note);
+            public void onPress(Note note) {
+                Log.d("!!!", "Note pressed " + simplePitchNotationFormatter.format(note));
+            }
+
+            @Override
+            public void onRelease(Note note) {
+                Log.d("!!!", "Note released " + simplePitchNotationFormatter.format(note));
             }
         });
         midiKeyboardDriver.open();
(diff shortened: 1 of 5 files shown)


piano-hero/mobile/src/main/java/com/novoda/pianohero/AndroidThingsActivity.java
```diff
@@ -7,19 +7,24 @@
...
-
         midiKeyboardDriver = new MidiKeyboardDriver.KeyStationMini32(this);
-        midiKeyboardDriver.attachListener(new MidiKeyboardDriver.KeyListener() {
+        midiKeyboardDriver.attachListener(new NoteListener() {
 
             @Override
-            public void onKeyPressed(MidiKeyboardDriver.Note note) {
-                Log.d("!!!", "note pressed " + note);
+            public void onPress(Note note) {
+                Log.d("!!!", "Note pressed " + simplePitchNotationFormatter.format(note));
```
ataulm: sharps _supported_ now :D

piano-hero/mobile/src/main/java/com/novoda/pianohero/MidiKeyboardDriver.java
```diff
@@ -1,263 +1,51 @@
...
 
-    interface KeyListener {
+    class KeyStationMini32 extends SimpleUsbMidiDriver implements MidiKeyboardDriver {
+        private NoteListener noteListener;
 
-        void onKeyPressed(Note note);
-
-    }
-
-    enum Note {
```
ataulm: the enum would be helpful for human comparison - since we're sticking to the key of C, formatting midi numbers to the relevant human readable note name is ezpz (SimplePitchNotationFormatter)

piano-hero/mobile/src/main/java/com/novoda/pianohero/NotesPlayedDispatcher.java
```diff
@@ -0,0 +1,31 @@
+package com.novoda.pianohero;
+
+import java.util.HashSet;
+import java.util.Set;
+
+class NotesPlayedDispatcher implements NoteListener {
```
ataulm: just pulled this up from the PianoActivity

piano-hero/mobile/src/main/java/com/novoda/pianohero/PianoC4ToB5Activity.java
```diff
@@ -29,7 +28,10 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
     @Override
     protected void onResume() {
         super.onResume();
-        inputView.attach(new NotesPlayedDispatcher(presenter));
+        NotesPlayedDispatcher noteListener = new NotesPlayedDispatcher(presenter);
+        midiKeyboardDriver.attachListener(noteListener);
+        midiKeyboardDriver.open();
+        inputView.attach(noteListener);
```
ataulm: using the same noteListener for both - you could use the onscreen keyboard or a midi input device.

the usb driver didn't seem to work when I connected my piano to my pixelc. Dunno if it's for more raw input, like on AndroidThings, whereas on PixelC android handles (and consumes?) the usb midi events and you have to register via the midi service.

PR #303, opened 2017-06-08
[MonkeyRunner] Remove the set of adb property from the child task

lgvalle:
## Scope
This PR removes adb path set in the monkey-runner config task.

The reason behind is:
1. It fails to find `command.adb` in certain project (let's say CCleaner)
2. Setting up `adb` path is done anyway in the parent task and it works.

## Paired with
@takecare


lgvalle:
I'm going to put this on hold for a moment, as it's not proven this change alone will solve the issue I found

PR #305, opened 2017-06-09
PT-464: Do not generate package-info when already available

tobiasheine:
[PT-464](https://novoda.atlassian.net/browse/PT-464)

**Scope of this PR**
We fixed a bug that caused the `package-info` to be generated, if the user manually put one into the source folder.

This was causing the build to fail, because there must be just one `package-info` per package.

**Tests**
We added a unit test which manually adds a `package-info` into the source folder and verifies there is none generated

Paired with @tasomaniac


gradle-nonnull-plugin/plugin/src/test/groovy/com/novoda/gradle/nonnull/AndroidProjectIntegrationTest.groovy
```diff
@@ -68,6 +67,24 @@ public class AndroidProjectIntegrationTest {
         assertThat(task.outcome).isEqualTo(TaskOutcome.SUCCESS)
     }
 
+    @Test
+    public void shouldNotGeneratePackInfoWhenItsAlreadyPartOfSources() {
```
Mecharyry: `shouldNotGeneratePackInfoWhenItsAlreadyPartOfSources` should be `shouldNotGeneratePackageInfoWhenItsAlreadyPartOfSources` to avoid the `Pack` abbreviation.

gradle-nonnull-plugin/plugin/src/main/groovy/com.novoda.gradle.nonnull/GeneratePackageAnnotationsTask.groovy
```diff
@@ -13,13 +12,14 @@ class GeneratePackageAnnotationsTask extends DefaultTask {
     Set<String> packages;
 
     @OutputDirectory
-    def outputDir
+    File outputDir
```
Mecharyry: `outputDirectory`, to avoid abbreviations where possible.

gradle-nonnull-plugin/plugin/src/test/groovy/com/novoda/gradle/nonnull/AndroidProjectIntegrationTest.groovy
```diff
@@ -68,6 +67,24 @@ public class AndroidProjectIntegrationTest {
...
+        def tempClass = new File(PROJECT.tempDir, 'package-info.java')
```
Mecharyry: `temporaryClass` to avoid abbreviations.

PR #306, opened 2017-06-14
[espresso-support] Add idling resource

Mecharyry:
### Problem 

View binding happens on a different thread to test execution, so tests aren't waiting for the binding to complete.

This can mean that test asserts are executed before binding is complete which is especially an issue when verifying method calls on a mock.

### Solution

Implement Espresso's IdlingResource so we can block while view binding takes place.

### Paired with
@ataulm

### DIFF
--- a/espresso-support/core/src/main/java/com/novoda/espresso/ViewTestRule.java
+++ b/espresso-support/core/src/main/java/com/novoda/espresso/ViewTestRule.java
@@ -45,7 +45,7 @@ public class ViewTestRule<T extends View> extends ActivityTestRule<ViewActivity>
 
     @Override
     public String getName() {
-        return "IdlingResourceName";
+        return ViewTestRule.class.getSimpleName() + "::IdlingResource";
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


espresso-support/core/src/main/java/com/novoda/espresso/ViewTestRule.java
```diff
@@ -25,11 +27,14 @@ protected Intent getActivityIntent() {
     }
 
     public void bindViewUsing(final Binder<T> binder) {
+        isIdle = false;
         getActivity().runOnUiThread(new Runnable() {
             @Override
             public void run() {
                 T view = getView();
                 binder.bind(view);
+                callback.onTransitionToIdle();
```
fourlastor: I don't necessarily think this is wrong but I think you'd have a simpler life if you delegated to [`CountingIdlingResource`](https://developer.android.com/reference/android/support/test/espresso/idling/CountingIdlingResource.html)

Mecharyry: Seems useful if you have multiple jobs occurring at the same time, we know that it will always be one so not sure if super useful atm...

ataulm: Thanks for the headsup @fourlastor we had a look and what we have is very simple. The counting one looks like a layer that wouldn't add anything here.

rock3r: 👍 let's ship then

PR #308, opened 2017-06-19
[espresso-support] Replace IdlingResource on ViewTestRule with Instrumentation.runOnMainSync

ataulm:
# Problem

Hastily implemented IdlingResource to ViewTestRule.bind method to make tests wait for the bind to complete before moving on.

Unfortunately, this doesn't stop lines in the tests from executing - I suppose there's specific methods that respect the idling resource.

# Solution

In any case, the binding of a view is not really a "resource". It turns out there's another method available on `Instrumentation` which allows you to run methods on the main thread (necessary if we want to bind views!) and here is the comment above:

```java
/**
 * Execute a call on the application's main thread, blocking until it is
 * complete.  Useful for doing things that are not thread-safe, such as
 * looking at or modifying the view hierarchy.
 * 
 * @param runner The code to run on the main thread.
 */
public void runOnMainSync(Runnable runner) {
    ...
}
```

We want to do that! This has the advantage over `activity.runOnUiThread(Runnable)` in that the test execution is blocked, so the test will wait for the view to be bound completely, even if the next line is just a `Log` statement.

Paired with @mr-archano (jk, he totes told me about this)

### DIFF
--- a/espresso-support/README.md
+++ b/espresso-support/README.md
@@ -30,16 +30,6 @@ public ViewTestRule<MovieItemView> rule = new ViewTestRule<>(R.layout.test_movie
 You can write BDD style tests here, highlighting the expected behaviour for your custom views, using a mixture of Espresso ViewActions and Mockito verifies:
 
 ```java
-@Before
-public void setUp() {
-    Espresso.registerIdlingResources(viewTestRule);
-}
-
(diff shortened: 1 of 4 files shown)


jackSzm:
LGTM !

PR #317, opened 2017-08-11
AndroidThings - Create an Inkyphat Driver

blundell:
Makes an AndroidThings version of this python library: https://github.com/pimoroni/inky-phat

Next Step : turn it into a contrib-driver and push back to Google.

The InkyPhat is an e-ink display which we can draw white, black or red pixels.
The image stays on the screen even if there is no power to the device.

This PR allows:
- individual pixels to be drawn 
- images to be converted and drawn
- text to be drawn
- simple shapes to be drawn (circles, squares, triangles)

|InkyPhat|Original|
|----|----|
|image|logo|

|InkyPhat|Original|
|----|----|
|image|ic_test_3|

|InkyPhat|Original|
|----|----|
|img-20170810-wa0013|ic_test_2|

|InkyPhat|Original|
|----|----|
|img-20170810-wa0021|img-20170810-wa0020|

|InkyPhat|Original|
|----|----|
|image|ic_test_1|

|InkyPhat|Original|
|----|----|
|img-20170810-wa0018|ic_test_4|

webcam-toy-photo14


InkyPhat/app/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    provided 'com.google.android.things:androidthings:0.4-devpreview'
```
xrigau: this could target `0.5` 👅

InkyPhat/app/src/main/java/com/novoda/inkyphat/ImageConverter.java
```diff
@@ -0,0 +1,94 @@
...
+        if (bitmapWidth < getOrientatedWidth() && bitmapHeight < getOrientatedHeight()) {
+            return sourceBitmap;
+        }
+
+        switch (scaleType) {
+            case FILL:
+                return imageScaler.fitXY(sourceBitmap, getOrientatedWidth(), getOrientatedHeight());
+            case START:
+            case CENTER:
+            case END:
```
xrigau: perhaps this should be 2 different things:

 1. `ScaleType` (fitXY, crop, etc.)
 2. `Alignment` (start, center, end)

InkyPhat/app/src/main/java/com/novoda/inkyphat/ShapeDrawer.java
```diff
@@ -0,0 +1,422 @@
+package com.novoda.inkyphat;
+
+import static java.lang.Math.cos;
+import static java.lang.Math.sin;
+
+public class ShapeDrawer {
```
xrigau: an alternative approach would be leveraging the drawing functions that already exist in the Android SDK's [`Canvas`](https://developer.android.com/reference/android/graphics/Canvas.html) backed by a `Bitmap` of the size of the inkyphat display and then you'd only need 1 method to draw a bitmap onto the inkyphat that way you'd avoid a lot of boilerplate which means potential of less bugs 💃

InkyPhat/app/build.gradle
blundell: will make another PR as I'd want to fully regression test :)

InkyPhat/app/src/main/java/com/novoda/inkyphat/ShapeDrawer.java
blundell: yeah so people could use `Inkphat.drawImage(Bitmap` you mean?

Having this class is extra you're right. Could potentially be removed (then in the demo's show shapes being drawn using Canvas).

I guess this class comes in handy if we implemented the "partial screen update" feature... leaning towards dropping it.

InkyPhat/app/src/main/java/com/novoda/inkyphat/InkyPhatTriColourDisplay.java
```diff
@@ -0,0 +1,188 @@
...
+        spiBus.write(data, data.length);
+    }
+
+    /**
+     * Wait for the e-paper driver to be ready to receive commands/data.
+     *
+     * @throws IOException error accessing GPIO
+     */
+    private void busyWait() throws IOException {
+        while (true) {
```
ouchadam: do you think it's worth having a timeout 1/2~ minutes to avoid a potential infinite loop?

ouchadam:
awesome work 💯

InkyPhat/app/src/main/java/com/novoda/inkyphat/InkyPhatTriColourDisplay.java
blundell: an infinite loop would mean that the hardware was broken though right? i.e. there isn't a recovery mechanism from it

ouchadam: or the code became out of sync and `busyWait` was called without resetting the pin its meant to be watching

blundell: the pin is always false except when the inkyphat is busy when it is true

blundell: i.e. the while loop always falls through, unless the gpio it is monitoring which is only controlled by the inkyphat is busy

A bit of me wants to add a timeout, but when i attempt to think logically its not needed

InkyPhat/app/src/main/java/com/novoda/inkyphat/ShapeDrawer.java
blundell: done! (see PR description) https://github.com/novoda/spikes/pull/317/commits/ea4a3a158faf5e98f232f0b8e877f2c83013eca1

InkyPhat/app/src/main/java/com/novoda/inkyphat/ImageConverter.java
blundell: https://github.com/novoda/spikes/pull/317/commits/07c70005cc4da2ca85ab4782bb82dfa6db69b468

xrigau: 👍

InkyPhat/app/src/main/java/com/novoda/inkyphat/ColorConverter.java
```diff
@@ -0,0 +1,34 @@
...
+import android.graphics.Color;
+
+class ColorConverter {
+
+    InkyPhat.Palette convertARBG888Color(int color) {
+        int red = Color.red(color);
+        int blue = Color.blue(color);
+        int green = Color.green(color);
+
+        if (red > 127 && blue > 127 && green > 127) {
```
xrigau: perhaps these could be configurable by the user

blundell: yeah, like an advanced feature. It's a trade off of configuration vs simplicity to use.

yeah if I made them a config in this class, then the factory could have another builder that took more options

PR #326, opened 2017-10-02
[espresso-support] Allow to create the View programatically

BraisGabin:
This allows to generate the view programatically and it doesn't introduce any BC.

This is usefull in some cases. For example, I'm using Flow so I instantiate all my Views by code. And I inject all the View dependecies before I add it to a parent. This way I know that my dependencies are warmed when `onAttachedToWindow` is called. With the previous code I couldn't test those views.

### DIFF
--- a/espresso-support/core/src/main/java/com/novoda/espresso/ViewTestRule.java
+++ b/espresso-support/core/src/main/java/com/novoda/espresso/ViewTestRule.java
@@ -1,11 +1,12 @@
 package com.novoda.espresso;
 
 import android.app.Instrumentation;
-import android.content.Intent;
 import android.support.annotation.LayoutRes;
 import android.support.test.InstrumentationRegistry;
 import android.support.test.rule.ActivityTestRule;
+import android.view.LayoutInflater;
 import android.view.View;
+import android.view.ViewGroup;
 
 public class ViewTestRule<T extends View> extends ActivityTestRule<ViewActivity> {
 
@@ -13,6 +14,7 @@ public class ViewTestRule<T extends View> extends ActivityTestRule<ViewActivity>
 
     @LayoutRes
     private final int id;
+    private T view;
 
     public ViewTestRule(@LayoutRes int id) {
         super(ViewActivity.class);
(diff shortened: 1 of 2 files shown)


hal9002:
Can one of the admins review this PR?

tasomaniac:
So, as far as I understand, you are just delaying the view inflation after the Activity is created/launched. I don't have any comment about the necessity of this delay but this much change seems not strictly necessary to me.

BraisGabin:
That delay is just a side effect. Do you know how to generate the view (by code) and set it to the activity before?

The important thing here is that you can generate the view by code or you can use the current way. Both ways work.

To illustrate a view that can't be tested right now:

```java
public class UserScreen extends FrameLayout {

    public static UserScreen createScreen(@NonNull Context context, ViewGroup parent, ActivityComponent component) {
        final UserScreen view = // create view
        component
            .plus(new UserModule(view, userId))
            .inject(view);
        return view;
    }

    @Inject
    Lifecycle lifecycle;

    @Inject
    Presenter presenter;

    @Override
    protected void onAttachedToWindow() {
        super.onAttachedToWindow();
        lifecycle.addObserver(presenter); // with the current @Rule this code crash here with NPE
    }

    @Override
    protected void onDetachedFromWindow() {
        super.onDetachedFromWindow();
        lifecycle.removeObserver(presenter);
    }
}
```

And how it can be tested using my PR:

```java
    @Rule
    public ViewTestRule<UserScreen> viewTestRule = new ViewTestRule<UserScreen>(-1) {
        @Override
        protected UserScreen createView(LayoutInflater inflater, ViewGroup parent) {
            return UserScreen.createScreen(inflater.getContext(), parent, activityComponentMock);
        }
    };
```

`createView` is in `ViewTestRule` as `getActivityIntent` is in `ActivityTestRule`.

ataulm:
@BraisGabin sorry for not seeing this before. I'll give it a look this week.

ataulm:
https://github.com/novoda/espresso-support/issues/1

ataulm:
We've discussed and migrated this issue to the new repo - https://github.com/novoda/espresso-support/pull/3

PR #334, opened 2018-01-02
[Alexa Space] Create cards with background

danybony:
Display questions using cards with background with a random space image on each one.
The images are from Pixbay and are all Free for commercial use without any need for attribution.
(the questions are still the ones related to reindeers, the final content will be added in a later PR)

image
image
image
image


danybony:
Yes, the problem with using the list is that we cannot have the question text in multiple lines, due to limitations to the template...

PR #350, opened 2018-01-11
[Gol] First iteration android client

tobiasheine:
**Description**
This PR ads the first fully functional iteration of the android client.

We had to downgrade to Gradle 3.3 since the `kotklin-platform-js` doesn't work yet with Gradle 4.

Additionally we renamed `BoardViewState` to `BoardInput` since it's the input of the board and added an extension function to `BoardView` in order to prevent code duplication among the different platforms.

Paired with @tasomaniac


tasomaniac:
No screenshots? :trollface:

PR #354, opened 2018-01-12
[GoL] Tests for business logic in common module

tasomaniac:
### Description

This PR introduces tests for Presenters and business logic of the game. 

- `GameLoop` is extracted as an interface to be able to have a stub implementation in the test. 
- `mockk` library is used because it is the only mocking framework with kotlin-multiplatform support. (I mean the first one we found)

PS: These tests were written first. We just didn't committed them until now. 

> Paired with @tobiasheine

### DIFF
--- a/game-of-life-multiplatform/common/src/main/kotlin/com/novoda/gol/presentation/AppModel.kt
+++ b/game-of-life-multiplatform/common/src/main/kotlin/com/novoda/gol/presentation/AppModel.kt
@@ -23,6 +23,9 @@ class AppModel {
     }
 
     fun selectPattern(pattern: PatternEntity) {
+        if (boardViewState.isIdle.not()) {
+            return
+        }
         boardViewState = boardViewState.copy(selectedPattern = pattern)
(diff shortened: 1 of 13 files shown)


game-of-life-multiplatform/common/src/main/kotlin/com/novoda/gol/presentation/AppModel.kt
```diff
@@ -23,6 +23,9 @@ class AppModel {
     }
 
     fun selectPattern(pattern: PatternEntity) {
+        if (boardViewState.isIdle.not()) {
```
tasomaniac: Ops.

game-of-life-multiplatform/web/src/main/js/index.js
```diff
@@ -1 +1 @@
-require("game-of-life-js.js");
\ No newline at end of file
+require("web.js");
```
tasomaniac: Ops :two:

PR #355, opened 2018-01-12
Fix lines tool to include .h files

amlcurran:
Should calculate the number of lines in header and implementation files for objective c


amlcurran:
If this isn't working, lets just close the PR. No-one uses this tool anyway.

PR #365, opened 2018-02-09
[PT-560]: Pattern Selection

tobiasheine:
### Description
Scope of this PR is to add the missing pattern selection to the kotlin native game of life client.

### Considerations
- remove not needed setup instruction to link the generated framework since it's a relative link
- align board and buttons correctly
- implement pattern selection using a `UIAlertController `
- bind `BoardPresenter` and `BoardView` in the `GameOfLifeViewController`

### Screenshots
pattern_selection


game-of-life-native/Game of Life/GameOfLifeViewController.swift
```diff
@@ -11,59 +11,104 @@ import Foundation
...
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        setupLayout()
+        appPresenter.bind(view: self)
+        boardPresenter.bind(boardView: boardView)
     }
-    
+
+    @objc func controlButtonAction(sender: UIButton!) {
+        let _ = onControlButtonClicked()
```
tobiasheine: Kotlin native compiles kotlin closures to a closure that returns `KGOLStdlibUnit`. To suppress the warning of the unused result we to assign it to `_`

game-of-life-native/Game of Life/views/UIBoard.swift
```diff
@@ -10,87 +10,70 @@ import UIKit
...
         let cellDimen = bounds.size.width / CGFloat(boardEntity.getWidth())
-        
-        for y in 0...boardEntity.getHeight()-1{
-            for x in 0...boardEntity.getWidth()-1{
-             let cell = boardEntity.cellAtPosition(x: x, y: y)
-                
-                let position = KGOLPositionEntity(x:x,y:y)
-                let cellView = UICell(frame: CGRect(x: CGFloat(x) * cellDimen, y: CGFloat(y) * cellDimen, width: cellDimen, height: cellDimen),position:position)
-                
+
```
tobiasheine: Just reformatting due to the switch from xcode to app code.

game-of-life-native/Game of Life/GameOfLifeViewController.swift
```diff
@@ -11,59 +11,104 @@ import Foundation
...
 
 class GameOfLifeViewController: UIViewController, KGOLAppView {
-    
-    private var appPresenter : KGOLAppPresenter?
-    private let controllButton = UIButton(frame: CGRect(x: 100, y: 100, width: 200, height: 50))
-    private let boardView = UIBoard(frame: CGRect(x: 0, y: 150, width: 300, height: 300))
-    
+
+    private let controlButton = UIButton(type: UIButtonType.system)
+    private let patternSelectionButton = UIButton(type: UIButtonType.system)
```
bertadevant: you do not really need to init the UIButton with system style, if you do not specify a style that one is the default. Also you can init it like this too `UIButton(type: .system)`

game-of-life-native/Game of Life/GameOfLifeViewController.swift
```diff
@@ -11,59 +11,104 @@ import Foundation
...
+                y: yOffset,
+                width: self.view.bounds.width / 2,
+                height: yOffset)
+
+        patternSelectionButton.frame = CGRect(
+                x: self.view.bounds.width / 2,
+                y: yOffset,
+                width: self.view.bounds.width / 2,
+                height: yOffset)
+        patternSelectionButton.setTitle("Choose a Pattern", for: .normal)
```
bertadevant: this should not be on a function called `setupLayout`

game-of-life-native/Game of Life/GameOfLifeViewController.swift
```diff
@@ -11,59 +11,104 @@ import Foundation
...
-            controllButton.addTarget(self, action: #selector(buttonAction), for: .touchUpInside)
-            self.view.addSubview(controllButton)
-            
-            boardView.backgroundColor = .blue
-            self.view.addSubview(boardView)
-        }
-        
-        appPresenter?.bind(view: self)
-        boardView.willAppear()
+        view.addSubview(controlButton)
```
bertadevant: might be worth it to create a function `setupViews()` that you call here and add the button `addTarget` and the functions for `backgroundColor` and `setTitle`

game-of-life-native/Game of Life/GameOfLifeViewController.swift
bertadevant: could Koitlin return a @discardable result?

game-of-life-native/Game of Life/GameOfLifeViewController.swift
```diff
@@ -11,59 +11,104 @@ import Foundation
...
+    @objc func controlButtonAction(sender: UIButton!) {
+        let _ = onControlButtonClicked()
+    }
+
+    @objc func patternSelectionButtonAction(sender: UIButton!) {
+        let alert = UIAlertController(title: "Patterns", message: "Please Choose a Pattern", preferredStyle: .actionSheet)
+
+        let patterns = KGOLPatternRepositoryCompanion().patterns() as NSArray
+        for pattern in patterns {
+            let patternEntity: KGOLPatternEntity = pattern as! KGOLPatternEntity
```
bertadevant: avoid force cast, it is preferable to use `guard let patternEntity: KGOLPatternEntity = pattern as? KGOLPatternEntity else { return }`

game-of-life-native/Game of Life/GameOfLifeViewController.swift
```diff
@@ -11,59 +11,104 @@ import Foundation
...
+
+            alert.addAction(UIAlertAction(title: patternEntity.getName(), style: .default, handler: { (action) in
+                let _ = self.onPatternSelected(patternEntity)
```
bertadevant: To avoid weird crashes on closures is always good to specify that the controller aka `self` might be nil sometimes (if it has been discarded while waiting for the closure to finish . 

```
handler: { [weak self] action in 
   let _ = self?.onPatternSelected(patternEntity) 
```
That way if the controller is nil when we try to do pattern selected it won't crash the app, it will just don't do anything

game-of-life-native/Game of Life/views/UIBoard.swift
```diff
@@ -10,87 +10,70 @@ import UIKit
...
+        for y in 0...boardEntity.getHeight() - 1 {
+            for x in 0...boardEntity.getWidth() - 1 {
```
bertadevant: would be more readable to do the for loops like this: `for x in 0..<boardEntity.getWidth()` than adding the `-1`

game-of-life-native/Game of Life/views/UIBoard.swift
```diff
@@ -10,87 +10,70 @@ import UIKit
...
         }
     }
-    
+
     @objc func onCellClicked(_ sender: UICell) {
         guard let position = sender.position else { return }
-        onCellClicked(position)
-    }
-    
-    func willAppear() {
```
bertadevant: why is this function gone now? we do not needed it anymore?

tobiasheine: It's just moved to the `GameOfLifeViewController`.

game-of-life-native/Game of Life/GameOfLifeViewController.swift
tobiasheine: Kotlin doesn't know about anything `discardable` since it's a swift feature.

bertadevant: that was my question, then your solution is the correct one! 👍

game-of-life-native/Game of Life/GameOfLifeViewController.swift
tobiasheine: If I dont specify it the text becomes invisible, so it doesn't seem to be the default for me 🤔

tobiasheine:
@bertadevant thx for the review, I addressed your comments 👍

game-of-life-native/Game of Life/GameOfLifeViewController.swift
```diff
@@ -11,59 +11,109 @@ import Foundation
...
+        boardPresenter = KGOLBoardPresenter(boardModel: model)
+
         super.init(coder: aDecoder)
     }
-    
+
+    private func setupLayout() {
+        let yOffset: CGFloat = CGFloat(35)
+
+        controlButton.frame = CGRect(
```
bertadevant: 👍

PR #368, opened 2018-02-26
[PT-580] Novoda Blog diversity dashboard

tobiasheine:
### Description
As part of this PR I used the `plugin-tester` to create a simple dashboard.

The dashboard analysis the [rss feed](https://www.novoda.com/blog/rss/) of our blog to:
- count `android`, `iOS` and `cross-platform` related tags based on a hardcoded list
- show the top 3 used tags

The downside of using the `rss feed` is that just the latest 15 entries are considered. An alternative would be to set up and consume the [ghost api](https://api.ghost.org/docs).

screen shot 2018-02-26 at 15 22 39


tobiasheine:
We agreed on moving this to the [dashboards repo](https://github.com/novoda/dashboard-plugins/pull/4)

PR #376, opened 2018-03-22
Android joystick

xrigau:
Adds an android (view-based) implementation of the joystick
And `set` the LED colours instead of `add` them (as this is what the Arduino code was doing)

Screenshot |
--- |
screenshot_1521737838 |


dungeon-crawler/app/build.gradle
```diff
@@ -4,7 +4,7 @@ android {
     compileSdkVersion 26
     defaultConfig {
         applicationId "com.novoda.dungeoncrawler"
-        minSdkVersion 26
+        minSdkVersion 24
```
blundell: ❌

dungeon-crawler/app/src/main/java/com/novoda/dungeoncrawler/JoystickView.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private final Runnable resetLastAction = () -> userAction = Action.NONE;
+
+    private void updateActionWith(MotionEvent motionEvent, Action thisAction) {
+        int action = motionEvent.getActionMasked();
+        if (action == MotionEvent.ACTION_DOWN) {
+            userAction = thisAction;
+        } else if (action == MotionEvent.ACTION_UP) {
+            delayResetAction(ACTION_MOVE_DELAY_MS);
```
xrigau: don't need this because the game polls for the `tilt` state in the joystick impl

PR #378, opened 2018-04-12
Android ARCore Sample

lgvalle:
## Scope
This PR adds an Android ARCore sample app. The sample is a rewrite of https://github.com/google-ar/arcore-android-sdk with a few tweaks:

* Code has been converted to Kotlin
* Render management has been moved away from the main activity into it's own class `NovodaSurfaceViewRenderer` which uses several collaborators to render and visualise each of the moving parts: background, points, planes and models.
* A view strip has been added at the bottom to display debug messages
* There are classes and files copied directly from the Google code. 
  * `com.novoda.spikes.arcore.google` : this contains the code to render different models using openGL and a helper to deal with taps on the screen
  * `/assets/models` : contains the model for Andy and the plane's grid
  * `/assets/shaders`: contains several shaders used during the rendering process

## Why this sample?
This sample is not introducing new functionality over the existing Google one. The aim is to, hopefully, make the code easier to understand and follow. 
Google's sample has all the rendering logic merged together in the same method [`MainActivity.onDrawFrame()`](https://github.com/google-ar/arcore-android-sdk/blob/master/samples/hello_ar_java/app/src/main/java/com/google/ar/core/examples/java/helloar/HelloArActivity.java#L242) which makes it really hard to follow

## Removed features

There are two main features removed from this sample to make it simpler:
* Orientation detection: the app is fixed in portrait
* Light adjustments: this is completely removed, as it's a nice extra but over complicates the example

## Screenshot

device-2018-04-12-103714


ARCore/app/src/main/java/com/novoda/spikes/arcore/MainActivity.kt
```diff
@@ -0,0 +1,87 @@
...
+            if (isARCoreInstalled && CameraPermissionHelper.isCameraPermissionGranted(this@MainActivity)) {
+                createOrResumeARSession()
+            } else {
+                showMessage(message)
+            }
+        }
+    }
+
+    private fun createOrResumeARSession() {
+        if (this::arSession.isInitialized.not()) {
```
xrigau: IMO this is a bit of a smell that the `arSession` should be a nullable `var` rather than a `lateinit var`, or maybe you want to use the `by lazy` initialiser if this just needs to happen once

ARCore/app/src/main/java/com/novoda/spikes/arcore/helper/ARCoreDependenciesHelper.kt
```diff
@@ -0,0 +1,37 @@
+package com.novoda.spikes.arcore.helper
+
+import android.app.Activity
+import com.google.ar.core.ArCoreApk
+import com.google.ar.core.exceptions.*
+
+object ARCoreDependenciesHelper {
+    class Result(val isARCoreInstalled: Boolean, val message: String)
```
xrigau: This is a great case for a `sealed class` 😋 

```
sealed class Result {
  class Installed : Result
  class NotInstalled : Result
}
```

lgvalle: Updated here 27b86a31

ARCore/app/src/main/java/com/novoda/spikes/arcore/MainActivity.kt
lgvalle: Updated here 01832ffb

ARCore/app/src/main/java/com/novoda/spikes/arcore/MainActivity.kt
```diff
@@ -0,0 +1,87 @@
...
+        checkAREnvironment()
+    }
+
+    private fun checkAREnvironment() {
+        ARCoreDependenciesHelper.isARCoreIsInstalled(this).apply {
+            if (this is ARCoreDependenciesHelper.Result.Failure) {
+                showMessage(message)
+            } else if (CameraPermissionHelper.isCameraPermissionGranted(this@MainActivity)) {
+                createOrResumeARSession()
+            }
```
lgvalle: With the introduction of a sealed class to model `Result` this is not looking so good ☝️  - Is there a better way to do that?

ARCore/app/src/main/java/com/novoda/spikes/arcore/MainActivity.kt
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    private fun createOrResumeARSession() {
```
xrigau: this should be renamed to just `resumeArSession`

ARCore/app/src/main/java/com/novoda/spikes/arcore/MainActivity.kt
xrigau: hmmm usually with `sealed` classes people tend to replace `if` with `when`
this code should almost compile 😅 : (also using static imports)

```kotlin
ARCoreDependenciesHelper.isARCoreIsInstalled(this).apply {
  when (this) {
    is Failure -> showMessage(message)
    is Success && CameraPermissionHelper.isCameraPermissionGranted(this@MainActivity) -> createOrResumeARSession()
  }
}
```

ARCore/app/src/main/java/com/novoda/spikes/arcore/MainActivity.kt
xrigau: Not relevant anymore

PR #381, opened 2018-05-01
Url proxy site for maintaining consistent friendly urls

ouchadam:
- A firebase function which reads the request path to match a key against a url in a firebase realtime database to redirect to.

- A react website to admin (add/remove) links

screen shot 2018-05-01 at 10 14 50

### DIFF
--- a/links/site/src/index.js
+++ b/links/site/src/index.js
@@ -55,11 +55,11 @@ class App extends React.Component {
 
 const AuthComponent = ({ }) => {
     return (
-        <button onClick={onSignInClicked()}> Sign in</button>
+        <button onClick={onSignInClicked}> Sign in</button>
     )
 }
 
-const onSignInClicked = () => () => {
+const onSignInClicked = () => {
     const provider = new firebase.auth.GoogleAuthProvider();
     firebase.auth().signInWithPopup(provider)
 }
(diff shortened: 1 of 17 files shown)


links/site/src/index.js
```diff
@@ -0,0 +1,237 @@
...
+
+}
+
+const AuthComponent = ({ }) => {
+    return (
+        <button onClick={onSignInClicked()}> Sign in</button>
+    )
+}
+
+const onSignInClicked = () => () => {
```
takecare: why `() => () =>`?

links/site/src/index.js
```diff
@@ -0,0 +1,237 @@
...
+    componentDidMount() {
+        this.unsubscribe = firebase.auth().onAuthStateChanged(user => {
+            this.setState({
+                isAuthed: Boolean(user) && !user.isAnonymous
+            })
+        })
+    }
+
+    componentWillUnmount() {
+        this.unsubscribe()
```
takecare: does it matter at this point to change the state?

ouchadam: react warns about memory leaks due to an async which may not have finished when unmounting

links/site/src/index.js
ouchadam:
```
        <button onClick={onSignInClicked()}> Sign in</button>
```

lol no idea, will remove the unneeded layer

PR #383, opened 2018-05-01
Arcore with tensorflow

Dorvaryn:
This adds tensorflow object recognition. The image data is sent to tensorflow and processed. The classified label output is used to search a model on poly and that model is added to the model collection. We can then select a model and drop it in the AR world.

screenshot_20180418-162621

This PR is for review. It will be merged into a new folder once reviewed to avoid making the original sample too complex.


ARCore/app/build.gradle
```diff
@@ -18,18 +18,26 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
+    aaptOptions {
+        noCompress "tflite"
```
Dorvaryn: If we compress the tensorflow models we cannot load them anymore

ARCore/app/src/main/java/com/novoda/spikes/arcore/ARClassifier.kt
```diff
@@ -0,0 +1,73 @@
...
+    private fun classifyImage(image: TextureReaderImage) {
+        if (frameModulo == 0) {
+            Thread(Runnable {
+                //use image.buffer
+                val label = classifier.classifyFrame(image)
+                Log.e(TAG, "Got result:\n$label")
+                listener.onObjectClassified(label)
+            }).start()
+        }
+        frameModulo = (frameModulo + 1) % 5
```
Dorvaryn: We only need to process every 5 frame

ARCore/app/src/main/java/com/novoda/spikes/arcore/ModelCollection.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    private val assets : List<PolyAsset>
+        get() = models.keys.toList()
+    
+    var currentModel: PolyAssetRenderer? = null
+
+    private var modelModulo = 0
+
+    override fun onObjectClassified(label: String) {
+        if (wasNotAlreadySearchedFor(label) && modelModulo == 0) { //We start with a basic modulo, we could improve with a reliability based load
```
Dorvaryn: We only query the API every 30 classifications  with the last classified label. We could improve this by running a total or using the confidence value back from tensorflow. A better version would query for the most likely object once it's confident enough maybe ?

Dorvaryn:
Discussion moved to https://github.com/novoda/spikes/pull/386

PR #387, opened 2018-05-02
hello torch iOS

lgvalle:
Sample app to make the device torch flash rapidly

Code is from this Stackoverflow question with some minor changes https://stackoverflow.com/questions/48956549/how-to-build-an-accurate-iphone-strobe-light-using-swift/48957759#48957759

In my tests it last for more than one hour without overheating

### DIFF
--- a/HelloTorchIOS/HelloTorchIOS/LightsLogic.swift
+++ b/HelloTorchIOS/HelloTorchIOS/LightsLogic.swift
@@ -1,0 +1,52 @@
+//
+//  LightsLogic.swift
+//  HelloTorchIOS
+//
+//  Created by Luis Valle on 08/05/2018.
+//  Copyright © 2018 Novoda. All rights reserved.
+//
+
+import Foundation
+import AVFoundation
+
+class LightsLogic {
+    
+    private let device: AVCaptureDevice? = AVCaptureDevice.default(for: AVMediaType.video)
+    
+    init() {
+        do {
+            try device?.lockForConfiguration()
+        }catch {
+            print("Torch could not be used")
+        }
+    }
+    
+    func toggleStrobe() {
+        if (device?.hasTorch == false) {
+            print( "torch unavailable")
+            return
+        }
+        
+        if isLightOn() {
+            turnLightOff()
+        } else {
+            turnLightOn()
+        }
+    }
+    
+    private func turnLightOn() {
+        do {
+            try device?.setTorchModeOn(level: 0.5)
+        }catch {
+            print("Torch could not be used")
+        }
+    }
+    
+    private func turnLightOff() {
+        device?.torchMode = .off
+    }
+    
+    private func isLightOn() -> Bool {
+        return device?.torchMode == .on
+    }
+}
--- a/HelloTorchIOS/HelloTorchIOS/LightsLogic.swift
+++ b/HelloTorchIOS/HelloTorchIOS/LightsLogic.swift
@@ -22,7 +22,7 @@ class LightsLogic {
     }
     
     func toggleStrobe() {
-        if (device?.hasTorch == false) {
+        guard device?.hasTorch == true else {
             print( "torch unavailable")
             return
         }
--- a/HelloTorchIOS/HelloTorchIOS/StrobeLights.swift
+++ b/HelloTorchIOS/HelloTorchIOS/StrobeLights.swift
@@ -2,73 +2,38 @@ import Foundation
 import AVFoundation
 
 class StrobeLights {
+    
+    let lightsController = LightsLogic()
     var counter: Int = 0
-    var timer: Timer
-    var isStrobing: Bool
-    var isLightOn: Bool
-    var frequency: Double
+    var timer: Timer = Timer()
+    var frequency: Double = 500
     var start = DispatchTime.now()
     var end = DispatchTime.now()
-    var active: Bool
-    var device: AVCaptureDevice
-    
-    init (){
-        self.counter = 0
-        self.timer = Timer()
-        self.isStrobing = false
-        self.isLightOn = false
-        self.frequency = 500
-        self.active = false
-        self.device = AVCaptureDevice.default(for: AVMediaType.video)!
-        do {
-            try device.lockForConfiguration()
-        }catch {
-            print("Torch could not be used")
-        }
-    }
+    var isCounting = false
     
     // Start Strobe process
     func toggleStrobe () {
-        if (device.hasTorch == false) {
-            print( "torch unavailable")
-            return
-        }
-        
-        print("Is light on: \(self.isLightOn)")
-        if (self.isLightOn == true) {
-            self.isLightOn = false
-            self.timer.invalidate()
+        if (self.isCounting == true) {
             print("Turning timer off")
+            self.isCounting = false
+            self.timer.invalidate()
+            
             self.end = DispatchTime.now()
             let nanoTime = end.uptimeNanoseconds - start.uptimeNanoseconds
             let timeInterval = Double(nanoTime) / 1_000_000_000
             print("I counted this high \(counter) in this many seconds \(timeInterval) ")
             counter = 0
-            incrementCounter()
         } else {
-            self.isLightOn = true
-            self.timer = Timer.scheduledTimer(timeInterval: 1/frequency, target: self, selector: #selector(incrementCounter), userInfo: nil, repeats: true)
             print("Turning timer on")
+            self.isCounting = true
+            self.timer = Timer.scheduledTimer(timeInterval: 1/frequency, target: self, selector: #selector(incrementCounter), userInfo: nil, repeats: true)
             self.start = DispatchTime.now()
         }
     }
     
     // Increase counter by one
     @objc func incrementCounter () {
-        self.toggleTorch(on: false)
+        self.lightsController.toggleStrobe()
         self.counter += 1
-        self.toggleTorch(on: true)
-    }
-    // Turns light on or off
-    func toggleTorch(on: Bool ) {
-        if on == true {
-            do {
-                try device.setTorchModeOn(level: 0.5)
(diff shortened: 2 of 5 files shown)


HelloTorchIOS/HelloTorchIOS/LightsLogic.swift
```diff
@@ -0,0 +1,52 @@
...
+    init() {
+        do {
+            try device?.lockForConfiguration()
+        }catch {
+            print("Torch could not be used")
+        }
+    }
+    
+    func toggleStrobe() {
+        if (device?.hasTorch == false) {
```
bertadevant: This could be a perfect use of a guard statement. A Guard statement on IOS is like an if/else but with the intent of returning out of the function if a condition is not met. 

```
guard device?.hasTorch else { 
  return 
} 
```

HelloTorchIOS/HelloTorchIOS/StrobeLights.swift
```diff
@@ -1,9 +1,39 @@
...
-//  Created by Luis Valle on 02/05/2018.
-//  Copyright © 2018 Novoda. All rights reserved.
-//
-
 import Foundation
+import AVFoundation
+
+class StrobeLights {
+    
+    let lightsController = LightsLogic()
```
bertadevant: Controller is reserved on IOS for `ViewControllers` might be better to rename this variable

HelloTorchIOS/HelloTorchIOS/StrobeLights.swift
```diff
@@ -1,9 +1,39 @@
...
-//  HelloTorchIOS
-//
-//  Created by Luis Valle on 02/05/2018.
-//  Copyright © 2018 Novoda. All rights reserved.
-//
-
 import Foundation
+import AVFoundation
+
+class StrobeLights {
```
bertadevant: Shouldn't this be renamed? Since it has nothing to do with the lights per se anymore, it just keeps a counter and timer right?

HelloTorchIOS/HelloTorchIOS/StrobeLights.swift
```diff
@@ -1,9 +1,39 @@
...
+    
+    let lightsController = LightsLogic()
+    var counter: Int = 0
+    var timer: Timer = Timer()
+    var frequency: Double = 500
+    var start = DispatchTime.now()
+    var end = DispatchTime.now()
+    var isCounting = false
```
bertadevant: If this variables should not be changed outside of the logic of the class I would set them to private

HelloTorchIOS/HelloTorchIOS/StrobeLights.swift
```diff
@@ -1,9 +1,39 @@
...
+    
+    // Start Strobe process
+    func toggleStrobe () {
+        if (self.isCounting == true) {
```
bertadevant: You do not need the `== true`, you can just do `if (isCounting) { }`

HelloTorchIOS/HelloTorchIOS/StrobeLights.swift
```diff
@@ -1,9 +1,39 @@
...
+            print("Turning timer off")
+            self.isCounting = false
+            self.timer.invalidate()
+            
+            self.end = DispatchTime.now()
+            let nanoTime = end.uptimeNanoseconds - start.uptimeNanoseconds
+            let timeInterval = Double(nanoTime) / 1_000_000_000
+            print("I counted this high \(counter) in this many seconds \(timeInterval) ")
+            counter = 0
```
bertadevant: I think this might be better abstracted into a private method in the same class, so you can read: 
```
if isCounting {
   turnTimerOff() 
   endDispatchTime() 
   resetCounter() 
}
``` 
And a similar abstraction for the else statement. How this strobe light works is a bit unclear since we have nit used it before so I think proper methods naming could help someone else pick up this code later on.

bertadevant: Also the part of the code of: 

```
let nanoTime = end.uptimeNanoseconds - start.uptimeNanoseconds
let timeInterval = Double(nanoTime) / 1_000_000_000
```

Could be set as a `print("I counted this high \(counter) in this many seconds \(timeInterval()) ")`. Working with time and date is tricky and separating it would clean up the code since it has nothing to do with the light strobe.

HelloTorchIOS/HelloTorchIOS/StrobeLights.swift
```diff
@@ -1,9 +1,39 @@
...
+            self.isCounting = true
+            self.timer = Timer.scheduledTimer(timeInterval: 1/frequency, target: self, selector: #selector(incrementCounter), userInfo: nil, repeats: true)
+            self.start = DispatchTime.now()
+        }
+    }
+    
+    // Increase counter by one
+    @objc func incrementCounter () {
+        self.lightsController.toggleStrobe()
+        self.counter += 1
```
bertadevant: You do not need to specify self here

HelloTorchIOS/HelloTorchIOS/ViewController.swift
```diff
@@ -10,16 +10,9 @@ import UIKit
 
 class ViewController: UIViewController {
 
-    override func viewDidLoad() {
-        super.viewDidLoad()
-        // Do any additional setup after loading the view, typically from a nib.
+    let strobeLights = StrobeLights()
+    
+    @IBAction func toggleTorch(sender: UIButton) {
+        strobeLights.toggleStrobe()
```
bertadevant: Since `toggleStrobe` has all the logic you could here directly call `incrementCounter` since is the method you call from timer. It would make more sense to have only one method access to the logic from outside instead of two.

HelloTorchIOS/HelloTorchIOS/StrobeLights.swift
lgvalle: I can tell you're loving the Clean Code book :)

HelloTorchIOS/HelloTorchIOS/ViewController.swift
lgvalle: I don't understand :/

HelloTorchIOS/HelloTorchIOS/StrobeLights.swift
bertadevant: I am loving it quite a lot 😄

HelloTorchIOS/HelloTorchIOS/LightsLogic.swift
```diff
@@ -22,7 +22,7 @@ class LightsLogic {
     }
     
     func toggleStrobe() {
-        if (device?.hasTorch == false) {
+        guard device?.hasTorch == true else {
```
bertadevant: You do not need `== true` It defaults to checking for true always.

HelloTorchIOS/HelloTorchIOS/LightsLogic.swift
```diff
@@ -0,0 +1,52 @@
...
+    init() {
+        do {
+            try device?.lockForConfiguration()
+        }catch {
+            print("Torch could not be used")
+        }
+    }
+    
+    func toggleStrobe() {
+        guard device?.hasTorch == true else {
```
bertadevant: No need to check for `== true` it already defaults to checking for true.

lgvalle: Not really, XCode complains because `hasTorch` seems to return an optional. If I remove the `== true` I get:
`Value of optional type 'Bool?' not unwrapped; did you mean to use '!' or '?'?`

bertadevant: No yeah you are right, if it is an optional yeah then it makes sense.

HelloTorchIOS/HelloTorchIOS/LightsLogic.swift
```diff
@@ -0,0 +1,52 @@
...
+//  Created by Luis Valle on 08/05/2018.
+//  Copyright © 2018 Novoda. All rights reserved.
+//
+
+import Foundation
+import AVFoundation
+
+class LightsLogic {
+    
+    private let device: AVCaptureDevice? = AVCaptureDevice.default(for: AVMediaType.video)
```
amlcurran: for future - if you make a failable initialiser, you could have removed this optionality throughout the class:

```swift
private let device: AVCaptureDevice

init?() {
   guard let device = AVCaptureDevice.default(for: AVMediaType.video) else {
      return nil
   }
   self.device = device
   ...
}
```

HelloTorchIOS/HelloTorchIOS/SwitchCounter.swift
```diff
@@ -0,0 +1,62 @@
...
+import AVFoundation
+
+class SwitchCounter {
+    
+    private let lightsLogic = LightsLogic()
+    private var counter: Int = 0
+    private var timer: Timer = Timer()
+    private var frequency: Double = 500
+    private var start = DispatchTime.now()
+    private var end = DispatchTime.now()
```
amlcurran: `DispatchTime` is a specific domain model for time in the "Grand Central Dispatch" class cluster (GCD is somewhat similar to `Executor` and `ExecutorService` + related classes). It's probably best you don't use it outside GCD APIs

You should probably instead just use `Date()`, which initialises a date set to now. There's also a bunch of helpful formatting tools around such as DateFormatter and DateIntervalFormatter. You can see the Date/Time class cluster here https://developer.apple.com/documentation/foundation/dates_and_times?language=swift

amlcurran: Here's some information about GCD https://developer.apple.com/documentation/DISPATCH

lgvalle:
It has already two 👍 - merging myself

PR #390, opened 2018-05-31
Immutability at its finest

blundell:
This PR attempts to make as much of the code immutable (through the use of `final` and upfront initialisation as possible. It moves related behaviours to the same place. Splits out separated concerns. And makes use of Java and Android as having more memory to play with than the arduino playground we came from.

From this we can deduce the list of "moving parts" for this program are:

```
GameEngine
	levelNumber
	previousFrameTime
	lastInputTime

	attackMillis
	attacking

	playerPosition
	stage
	stageStartTime
	lives

	millis

Boss
	alive
	position
	lives
	
Enemy
	alive
	position
	
EnemySpawn
	alive
	lastSpawned
	
Lava
	state
	lastOn
	
```


dungeon-crawler/game_engine/src/main/java/com/novoda/dungeoncrawler/Conveyor.java
```diff
@@ -3,40 +3,42 @@
 import static com.novoda.dungeoncrawler.Direction.RIGHT_TO_LEFT;
 
 class Conveyor {
-    private boolean alive;
```
blundell: `alive` had dual meanings in the codebase

1: as we expected if the enemy was dead or alive

2: things that where dead could be overriden in the array pool of enemies, we now avoid that and allocate all the space we need upfront

dungeon-crawler/game_engine/src/main/java/com/novoda/dungeoncrawler/EnemySpawner.java
```diff
@@ -0,0 +1,61 @@
+package com.novoda.dungeoncrawler;
+
+class EnemySpawner {
+    private final long activate;
+    private final long rate;
+    private final int position;
+    private final int direction;
+    private final int speed;
+
+    private long lastSpawned;
```
xrigau: 😱

PR #401, opened 2019-01-14
Multiplatform Movies - Popular Movies Gallery 

tobiasheine:
**Description**
This is the first PR for our kotlin multiplatform movies app. The first screen will be a gallery rendering a grid of movies from `https://api.themoviedb.org/3/movie/popular`. This adds the business logic and presentation layer consuming the endpoint and passing the data up to the `GalleryActivity` and `GalleryViewController`.

In this iteration, we just render the number of received movie posters. The next step will be implementing the native UI for the grid.

**Considerations**
The tests are located in `jvmTest` although they don't test a specific `jvm` implementation.  Instead, they should be located in `commonTest`. There are two reasons for this:
- There is no mocking framework supporting `Kotlin/Native` due to the lack of missing reflection. So I am using `Mockito`.
- Testing the coroutines is difficult with `Kotlin/Native` as `runBlocking` does not _yet_ exist. The reason is that coroutines are still single threaded.

But imho that's fine as we're not testing any `Kotlin/Native` implementation.

screen shot 2019-01-14 at 18 20 41


multiplatform-movies/common/src/commonMain/kotlin/com.novoda.movies.core/NetworkingDependencyProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+
+class NetworkingDependencyProvider {
+
+    fun provideAuthenticatedClient() = HttpClient {
+        authenticatedRequestsEngine()
+    }
+
+    private fun HttpClientConfig<*>.authenticatedRequestsEngine() {
+        engine {
+            install(KEY_AUTH_INTERCEPTOR) {
```
tobiasheine: Using this api we can easily share the configuration for the ktor http client, in this case we want to add the `api_key` to each request.

multiplatform-movies/common/src/commonMain/kotlin/com.novoda.movies.gallery/GalleryPresenter.kt
```diff
@@ -0,0 +1,37 @@
+package com.novoda.movies.gallery
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import kotlin.coroutines.CoroutineContext
+
+class GalleryPresenter internal constructor(private val uiContext: CoroutineContext, private val galleryFetcher: GalleryFetcher) : CoroutineScope {
```
tobiasheine: Similar to `schedulers` we inject the `CoroutinesContext` to have control over the theading.  For production we use `UI` wherein tests it's `Unconfined` (similar to trampoline).

multiplatform-movies/common/src/commonMain/kotlin/com.novoda.movies.core/NetworkingDependencyProvider.kt
charroch: Can this not come from the build process?

tobiasheine: Yes! Next step is making out build properties plugin multi platform aware : 💯

multiplatform-movies/common/src/commonMain/kotlin/com.novoda.movies.gallery/GalleryPresenter.kt
```diff
@@ -0,0 +1,37 @@
...
+
+
+    fun startPresenting(view: View) {
+        job = Job()
+
+        launch {
+            try {
+                val gallery = galleryFetcher.fetchGallery()
+                view.render(gallery)
+            } catch (exception: Exception) {
```
florianmski: Did you had a look at https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-exception-handler/index.html to handle Exceptions in coroutines?

tobiasheine: nice one 👍 
10fa827c

PR #403, opened 2019-03-21
[Multiplatform Movies] - Render movie posters in android client

Michal-Novoda:
# What does this PR do?
* Implementation of rendering of movie posters in android client

# Implementation Considerations
* Render movie posters in a `RecyclerView` using an `Adapter`.
* Download and render images using `Glide` image loading library.

# Design Review, Screenshots, Gifs
Screenshot_1553188256

# Has the solution been tested?
* no tests added since this is just the UI part

# Paired with
* @tobiasheine


multiplatform-movies/android/src/main/java/com/novoda/movies/gallery/GalleryAdapter.kt
```diff
@@ -0,0 +1,45 @@
...
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import com.bumptech.glide.Glide
+import com.novoda.movies.R
+
+
+class GalleryAdapter(private val moviePosterWidthInPixels: Int) : RecyclerView.Adapter<GalleryItemViewHolder>() {
+
+    private var moviePosters = listOf<MoviePoster>()
```
florianmski: Any reason why this is not a `val`?

multiplatform-movies/android/src/main/java/com/novoda/movies/gallery/GalleryActivity.kt
```diff
@@ -2,16 +2,26 @@ package com.novoda.movies.gallery
 
 import android.os.Bundle
 import android.support.v7.app.AppCompatActivity
+import android.support.v7.widget.GridLayoutManager
+import android.widget.Toast
 import com.novoda.movies.R
 import kotlinx.android.synthetic.main.activity_gallery.*
 
+private const val NO_OF_COLUMNS = 3
```
florianmski: `NO` or `NB`?

multiplatform-movies/android/src/main/java/com/novoda/movies/gallery/GalleryActivity.kt
```diff
@@ -25,10 +35,12 @@ class GalleryActivity : AppCompatActivity(), GalleryPresenter.View {
...
-        label.text = message ?: getString(R.string.gallery_error)
+        Toast.makeText(this, message, Toast.LENGTH_LONG).show()
     }
 
     override fun render(gallery: Gallery) {
-        label.text = resources.getString(R.string.gallery_items, gallery.moviePosters.size)
+        galleryAdapter.updateWith(gallery)
     }
+
+    private fun moviePosterWidthInPixels(): Int = resources.displayMetrics.widthPixels / NO_OF_COLUMNS
```
florianmski: The result is truncated (fractional part thrown away) and not rounded to the closest integer when dividing two ints.
Might be fine here because the number will be high but if we need something pixel perfect we probably need to have a float somewhere and then round the result.

multiplatform-movies/android/src/main/java/com/novoda/movies/gallery/GalleryAdapter.kt
```diff
@@ -0,0 +1,45 @@
...
+import android.support.v7.widget.RecyclerView
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import com.bumptech.glide.Glide
+import com.novoda.movies.R
+
+
+class GalleryAdapter(private val moviePosterWidthInPixels: Int) : RecyclerView.Adapter<GalleryItemViewHolder>() {
```
florianmski: I think nowadays we should be able to avoid having to have pixels measurements for images and grid to play well.
Usually tweaking the constraintlayout should work but I'm not 💯 % sure of the exact solution

multiplatform-movies/android/src/main/res/layout/activity_gallery.xml
```diff
@@ -5,12 +5,12 @@
...
+        tools:context=".gallery.GalleryActivity">
 
-    <TextView
-            android:id="@+id/label"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
+    <android.support.v7.widget.RecyclerView
+            android:id="@+id/movie_collection"
+            android:layout_width="0dp"
+            android:layout_height="0dp"
```
florianmski: we usually have a `match_constraint` dimension we use for that

multiplatform-movies/android/src/main/java/com/novoda/movies/gallery/GalleryAdapter.kt
gbasile: they replace the entire array when the adapter is updated

florianmski: yes but we usually `clear` and `addAll` which avoids creating a new list.
Not a biggy though

multiplatform-movies/android/src/main/java/com/novoda/movies/gallery/GalleryAdapter.kt
amlcurran: you could also override (or use [GridLayoutManager](https://developer.android.com/reference/android/support/v7/widget/GridLayoutManager))

multiplatform-movies/common/src/commonMain/kotlin/com.novoda.movies.gallery/Gallery.kt
```diff
@@ -1,7 +1,5 @@
 package com.novoda.movies.gallery
 
-import io.ktor.http.Url
-
 data class Gallery(val moviePosters: List<MoviePoster>)
 
-data class MoviePoster(val movieId: Long, val thumbnailUrl: Url)
+data class MoviePoster(val movieId: Long, val thumbnailUrl: String)
```
amlcurran: using a primitive here is a bit codesmell - I see you append it later, in which case this is not a thumbnailUrl but a thumbnailUrlFragment, or suffix or something.

multiplatform-movies/android/src/main/java/com/novoda/movies/gallery/GalleryAdapter.kt
amlcurran: oh you do. Then I don't think you need to modify the layout params of the view holder manually?

tobiasheine: Unfortunately, we still have to do that as the default behavior of the `GridLayoutManager` is not to fill the space automatically :(

multiplatform-movies/android/src/main/java/com/novoda/movies/gallery/GalleryActivity.kt
Michal-Novoda: `No.` Seems to be the correct abbreviation for number : 
https://english.stackexchange.com/questions/43095/what-is-the-correct-abbreviation-for-the-word-numbers

florianmski: Interesting, always thought it was NB, maybe in french?
👍

multiplatform-movies/common/src/commonMain/kotlin/com.novoda.movies.gallery/Gallery.kt
tobiasheine: It is the full url, the fragmet is the one in the `com.novoda.movies.gallery/ApiGallery.kt:12`

PR #425, opened 2019-10-02
[💻 ⤸ 🧹] CRUD

florianmski:
Add a bunch of new routes as part of our CRUD session


microservices-swift/microservices-swift.xcodeproj/xcshareddata/xcschemes/microservices-swift-Package.xcscheme
```diff
@@ -53,6 +53,8 @@
             </BuildableReference>
```
gbasile: these files should all be ignored in git :/

PR #436, opened 2020-02-17
BOT/Project Update

Mecharyry:
## Problem 
`TpBot` is super out of date 😬 but I want to get something working so that we can eventually attempt to hook it up to `Peepz` if possible.

## Solution
Update to `androidX` dependencies, update all dependencies including dagger 😄 

## Next Steps
- Add a GitHub Actions against the `tpbot-feature-branch` so we can report test coverage etc. 
- I have a branch that does some work fixing the `MovementService`, the one that passes commands to the hardware.
- I'm thinking about moving this over to `kotlin` as well 😄


danybony:
And yeah looking forward to see this in kotlin too 😅

PR #437, opened 2020-02-17
BOT/Fix Movement Service

Mecharyry:
## Problem 
😂 `Commits on Oct 11, 2017` 😂 

My memory might be a bit hazy on this one 😂 but I remember we were seeing a few crashes coming from the `MovementService` because it uses some executors to perform actions. Some of these actions would occur when navigating back or minimising the app, which would lead to a crash because it would attempt to call something that had already been destroyed 😬 

## Solution

I believe I tried to fix this by introducing a layer around the `Executor` that would attempt to remove actions when they were no longer necessary i.e. when minimising the app. This abstraction, at the time, was useful because it hide the execution behind an interface which allows testing and swapping out if we wanted to.


TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/threading/MyLooperDelayedExecutor.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class MyLooperDelayedExecutor implements Executor.DelayedExecutor {
+
+    private final Handler handler;
+    private final Map<Action, Runnable> runnables;
+
+    public static MyLooperDelayedExecutor newInstance() {
+        Handler handler = new Handler();
```
JozefCeluch: Should this executor specify the `Looper` too?

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/device/usb/UsbDeviceConnection.java
```diff
@@ -71,7 +72,7 @@ public void onReceive(String data) {
 
     @Override
     public void send(String data) {
-        // TODO: send data to serial port.
+        serialPortMonitor.tryToSendCommand(data);
```
danybony: What? how were we even sending commands? 😅

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/threading/MyLooperDelayedExecutor.java
Mecharyry: Probably 🤔 I'll take a look 😄

TelepresenceBot/mobile/src/main/java/com/novoda/tpbot/bot/device/usb/UsbDeviceConnection.java
Mecharyry: Yea, this was the thing I found weird because I remember seeing a video where you moved the bot 😂

Mecharyry: I spent ages wondering why it wasn't working, only to stumble across the TODO 😂

danybony: Lol who knows how I made it working... oh well this makes sense

